Issue #2, opened 2023-04-24
Finalize documentation for provider side


ethanfrey:
Provider clear description of all contract roles.
Provide API definitions as Rust files

PR #17, opened 2023-04-27
Define producer APIs

### LINKED ISSUE #3: Define all contract interfaces / APIs as Rust Files
Make a new package for the shared types.
Define messages and associated types for all contracts on the provider side.

### LINKED ISSUE #4: Provide contract stubs
Create basic contracts in the [Sylvia](https://cosmwasm.github.io/sylvia-book/index.html) framework.
Define all provider-side contracts and generic interfaces.

ethanfrey:
Part of #2 
Closes #3
Closes #4 

Provide Rust APIs for how the contracts interact


contracts/native-staking/src/contract.rs
```diff
@@ -0,0 +1,120 @@
...
+
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub struct NativeStakingContract<'a> {
+    // TODO
+    config: Item<'a, Config>,
+}
+
+#[contract(error=ContractError)]
```
hashedone, line 20: Just be aware that we would probably change it to be:

```
#[contract]
#[error(ContractError)]
```

For now it is how it should be, but there are reasons to break this one.

contracts/native-staking/src/contract.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// unstakes the given amount from the given validator on behalf of the calling user.
+    /// returns an error if the user doesn't have such stake.
+    /// after unbonding period, it will allow the user to claim the tokens (returning to vault)
+    #[msg(exec)]
+    fn unstake(
+        &self,
+        _ctx: ExecCtx,
+        _validator: String,
+        _amount: Uint128,
```
hashedone, line 49: This is wrong - this generates a message with fields `_validater` and `_amount`. We left it this way for two reasons:
1. Parsing this and forcefully stripping the `_` from before the name would disallow creating a message with a field starting with `_`, which technically might be a need I think. We were not fully sure about that.
2. We decided it is actually really improbable to have unused fields on final messages - especially since this problem does not occur when working with traits implementations.

We were thinking about adding a warning about this behaviour generating the code and adding a way to silence it while maintaining `_` in the name, but we are not sure what way is the best. Right now, the solution is to add `#[allow(unused)]` - either for the unused fields one by one or before the entire function, which is under development.

For obvious reasons - the problem does not occur for context, and it can have underscore.

contracts/native-staking/src/types.rs
```diff
@@ -0,0 +1,35 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::Addr;
+
+/*** state ***/
```
hashedone, line 4: I would personally split this `types.rs` into `state.rs` (for state types) and `msg.rs` (for message-related-types - especially those are actual response messages).

contracts/vault/src/contract.rs
hashedone, line 1: In general - all the same comments as in prev file.

hashedone (commented):
Mostly ok, but generated messages would contain obsolete `_` signed before names.

.gitignore
```diff
@@ -1,28 +1,20 @@
...
-# Auto-gen
-.cargo-ok
+schema/
 
-# Build artifacts
-*.wasm
-hash.txt
-contracts.txt
-artifacts/
+.DS_Store
```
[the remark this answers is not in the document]
ethanfrey: This is the same as many other repos, like cw-plus.
You have valid points, but I just kind of stick to those standards (and if we change, we change everywhere)

Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[workspace]
+members = ["packages/*", "contracts/*"]
+
```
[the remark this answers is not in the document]
ethanfrey: Maybe that was the issue with my mt problem?

contracts/native-staking/src/types.rs
ethanfrey: I kind of agree.. but when I see lots of files of < 30 lines, it seems easier to join.
Will split

contracts/native-staking/src/contract.rs
ethanfrey: I used `let _ = (a, b, c)` for the args. I don't like `#[allow(unused)]` as that helps me sometimes. But yeah, will make the proper names in args.

contracts/native-staking/src/contract.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        // parse message to find validator to stake on
+        let StakeMsg { validator } = from_slice(&msg)?;
+        let _ = validator;
+
+        // look up if there is a proxy to match
+        // instantiate or call stake on existing
+        todo!();
+    }
+
```
JakeHartnell, line 85: Kind of want a query to look up my proxy contract if it exists.

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+    #[msg(query)]
+    fn config(&self, _ctx: QueryCtx) -> Result<ConfigResponse, ContractError> {
+        todo!()
+    }
+
+    /// Returns all pending unbonding for this account
+    /// TODO: can we do that with contract API?
+    /// Or better they use cosmjs native delegation queries with this proxy address
+    #[msg(query)]
+    fn unbonding(&self, _ctx: QueryCtx, _account: String) -> Result<ClaimsResponse, ContractError> {
```
JakeHartnell, line 164: Does this need the `account` argument? Only one per contract?

contracts/vault/src/contract.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[msg(instantiate)]
+    pub fn instantiate(
+        &self,
+        ctx: InstantiateCtx,
+        denom: String,
+        local_staking: StakingInitInfo,
+    ) -> Result<Response, ContractError> {
+        let config = Config {
+            denom,
+            // We set this in reply, so proper once the
```
JakeHartnell, line 45:
```suggestion
            // We set this in reply, so proper once the reply message completes successfully
```

contracts/native-staking/src/contract.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Returns the maximum percentage that can be slashed
+    /// TODO: any way to query this from the chain? or we just pass in InstantiateMsg???
```
JakeHartnell, line 87: Could probably do this with Stargate queries? A bit torn on what to do here... IMO querying it could help reduce edge cases. Perhaps we could make it an enum and have a struct that can handle the query logic?

Just spit-balling here...

```rust
pub enum MaxSlashPercentage {
    /// A fixed value, this cannot be changed for the life of the contract
    Hardcoded  { max_slash: Decimal },
   /// Queries the max percentage that can be slashed via Stargate
    StargateQuery { },
}

pub struct MaxSlash {
    max_slash: MaxSlashPercentage
}

impl MaxSlash {
    fn get_max_slash_percentage (&self, ctx: ExecCtx) -> Result<Decimal, ContractError> {
        match self.max_slash {
            // If hardcoded we just return it
            MaxSlashPercentage::HardCoded { max_slash } => Ok(max_slash)
            // If not we do the stargate query
            MaxSlashPercentage::StargateQuery {} => {
                 // do stargate query and return value
            }
        }
    }
}
```

My hesitation with Stargate Queries is that they wouldn't work for chains using CosmWasm but not the SDK...

contracts/vault/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+#[cw_serde]
+pub struct ConfigResponse {
+    pub denom: String,
+
+    /// The address of the local staking contract (where actual tokens go)
+    pub local_staking: String,
+}
+
+#[cw_serde]
+pub struct BalanceResponse {
```
JakeHartnell, line 51: Nit on name, but this is returning more info than just a balance. Maybe `AccountResponse` and `QueryMsg::Account`?

packages/apis/src/local_staking_api.rs
```diff
@@ -0,0 +1,39 @@
...
+use sylvia::types::{ExecCtx, QueryCtx};
+use sylvia::{interface, schemars};
+
+#[cw_serde]
+pub struct MaxSlashResponse {
+    pub max_slash: Decimal,
+}
+
+// TODO: question - staking should know which is vault, vault should know what is local staking...
+// How to best handle the chicken and egg problem (2 step init with Option?)
```
JakeHartnell, line 12: I think you solved this? We instantiate the local staking contract when we create the vault?
```suggestion
```

packages/apis/src/vault_api.rs
```diff
@@ -0,0 +1,34 @@
+use cosmwasm_std::{Response, StdError, Uint128};
+use sylvia::types::ExecCtx;
+use sylvia::{interface, schemars};
+
+// TODO: question - local staking should know which is vault, vault should know what is local staking...
+// How to best handle the chicken and egg problem (2 step init with Option?)
+
```
JakeHartnell, line 7: Duplicate comment.
```suggestion
```

JakeHartnell (approved):
Few minor comments, but this is the right direction from my perspective.

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+
+#[contract(error=ContractError)]
+impl NativeStakingProxyContract<'_> {
+    pub const fn new() -> Self {
+        Self {
+            config: Item::new("config"),
+        }
+    }
+
+    /// The caller of the instantiation will be the native-staking contract.
```
maurolacy, line 27: According to the Architecture doc, the caller will be the vault contract. Can you clarify?

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// We stake `funds.info` on the given validator
+    #[msg(instantiate)]
+    pub fn instantiate(
+        &self,
+        ctx: InstantiateCtx,
+        denom: String,
+        owner: String,
```
maurolacy, line 34: Does this imply one proxy contract per user / address? Is this `owner` more like an admin (the valut contract)?

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /// The caller of the instantiation will be the native-staking contract.
+    /// We stake `funds.info` on the given validator
+    #[msg(instantiate)]
+    pub fn instantiate(
+        &self,
+        ctx: InstantiateCtx,
+        denom: String,
```
maurolacy, line 33: Is this the `denom` on the provider side, or the receiver side? I understand it's in the receiver's, and that this will then determine the respective consumer on the other chain.

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+        owner: String,
+        validator: String,
```
maurolacy, line 35: This is needed because of auto-staking, below. I think it makes sense for robustness, that any initially sent funds here are already being staked. As this is the entire purpose of this impl.

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+        };
+        self.config.save(ctx.deps.storage, &config)?;
+        set_contract_version(ctx.deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+        // stake info.funds on validator
+        let res = self.stake(ctx, validator)?;
+
+        // set owner as recipient of future withdrawls
+        let set_withdrawl = DistributionMsg::SetWithdrawAddress {
+            address: config.parent.into_string(),
```
maurolacy, line 50:
```suggestion
            address: config.owner.into_string(),
```

?

If I understand this properly:

 - `sender` would (typically) be the vault contract, and becomes `parent` here.
 - `owner` would be (typically) set by the vault to the original user, the provider of the funds.

In principle it should be possible to instantiate one of these contracts directly, and that is not a problem. The vault would be just a convenient way to do it, AFAIK.

contracts/native-staking/src/contract.rs
maurolacy: Same question. `account` seems wrong here, as this proxy handles only one address (that of the `owner`) over only one chain (that corresponding to `denom`).

maurolacy (commented):
Looks good. Reviewed the `native-staking-proxy` contract. Will review the rest in another iteration.

contracts/native-staking-proxy/src/contract.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// stakes the tokens from `info.funds` to the given validator.
+    /// can only be called by the parent contract.
+    #[msg(exec)]
+    fn stake(&self, ctx: ExecCtx, _validator: String) -> Result<Response, ContractError> {
+        let cfg = self.config.load(ctx.deps.storage)?;
+        ensure_eq!(cfg.parent, ctx.info.sender, ContractError::Unauthorized {});
+
+        todo!()
```
maurolacy, line 62: IBC call to receiver side happens here (commenting just for clarity / confirmation).

contracts/native-staking-proxy/README.md
```diff
@@ -0,0 +1,17 @@
+# Native Staking Proxy
```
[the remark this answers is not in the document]
JakeHartnell: This is not yet reflected in that doc. We should probably update the flow charts to include it.

contracts/native-staking-proxy/src/contract.rs
JakeHartnell: Ok, so the way this works is: `vault -> native-staking -> native-staking-proxy`.

The user story behind this is simple: *as a user, I want to cross-stake my native asset as well as stake it to secure the native chain.*

Why do we have the proxy? Because we want the user to be able to manage their native-stake just like they would normally (with the ability to vote, etc.). If all assets are sitting in one giant contract, this becomes difficult.

contracts/native-staking-proxy/src/contract.rs
JakeHartnell: Yes. One proxy contract per user.

Flow:
1. I deposit 1 $Juno into the vault
2. I decided to stake it to secure the Juno chain (not cross-staking yet)
3. A proxy account that I control is created for me where by I will be able to manage my stake

contracts/native-staking-proxy/src/contract.rs
JakeHartnell: We're not doing `cross-staking` yet. This is *local staking*.

We are building out a piece that was missing in the prototype, staking tokens locally *as well as* cross-staking them. Importantly, both are integrated into the vault.

contracts/native-staking-proxy/src/contract.rs
JakeHartnell: You are right.

contracts/native-staking-proxy/src/contract.rs
JakeHartnell: The caller of this proxy contract will be the native staking contract.

contracts/native-staking-proxy/src/contract.rs
JakeHartnell: No, this is local only.

JakeHartnell: Provider contracts coming soon.

contracts/native-staking-proxy/src/contract.rs
maurolacy: I see. It's called 'proxy' because it's an agent that acts on behalf of the user for some actions like unbonding, etc.

Thanks for the clarifications.

contracts/vault/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+/**** api ****/
+
+/// This is the info used to construct the native staking contract
+#[cw_serde]
+pub struct StakingInitInfo {
+    /// Admin for the local staking contract. If empty, it is immutable
+    pub admin: Option<String>,
+    /// Code id used to instantiate the local staking contract
+    pub code_id: u64,
+    /// msg is the JSON-encoded InstantiateMsg struct (as raw Binary)
```
maurolacy, line 36:
```suggestion
    /// JSON-encoded local staking `InstantiateMsg` struct (as raw `Binary`)
```

contracts/vault/src/error.rs
```diff
@@ -0,0 +1,33 @@
...
+    Unauthorized {},
+
+    #[error("Claim is locked, only {0} can be unbonded")]
+    ClaimsLocked(Uint128),
+
+    #[error("The address doesn't have sufficient balance for this operation")]
+    InsufficentBalance,
+
+    #[error("The leinholder doesn't have any claims")]
+    UnknownLeinholder,
```
maurolacy, line 26: Here and elsewhere, do you mean 'lien'?

```suggestion
    #[error("The lien holder doesn't have any claims")]
    UnknownLienholder,
```

contracts/vault/src/error.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    #[error("The leinholder doesn't have enough claims for the action")]
+    InsufficientLein,
```
maurolacy, line 29:
```suggestion
    #[error("The lien holder doesn't have enough claims for the action")]
    InsufficientLien,
```

contracts/vault/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[cw_serde]
+pub struct Balance {
+    pub bonded: Uint128,
+    pub claims: Vec<LeinAddr>,
+}
+
+#[cw_serde]
+pub struct LeinAddr {
+    pub leinholder: Addr,
```
maurolacy, line 23:
```suggestion
pub struct LienAddr {
    pub lien_holder: Addr,
```

contracts/vault/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+/// This is the info used to construct the native staking contract
+#[cw_serde]
+pub struct StakingInitInfo {
+    /// Admin for the local staking contract. If empty, it is immutable
+    pub admin: Option<String>,
+    /// Code id used to instantiate the local staking contract
+    pub code_id: u64,
+    /// msg is the JSON-encoded InstantiateMsg struct (as raw Binary)
+    pub msg: Binary,
+    /// A human-readable label for the contract (will use a default if not provided)
```
maurolacy, line 38:
```suggestion
    /// A human-readable label for the local staking contract (will use a default if not provided)
```

contracts/vault/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+#[cw_serde]
+pub struct BalanceResponse {
+    pub bonded: Uint128,
+    pub free: Uint128,
+    pub claims: Vec<Lein>,
+}
+
+#[cw_serde]
+pub struct Lein {
+    pub leinholder: String,
```
maurolacy, line 59:
```suggestion
pub struct Lien {
    pub lien_holder: String,
```

maurolacy (approved):
LGTM. Only minor syntax / naming stuff.

contracts/vault/src/error.rs
JakeHartnell: Seems `leinholder` is an actual word from a quick search...

maurolacy: Hmm, it seems to me that the right or more common expression is '[lien](https://www.thefreedictionary.com/Lein) holder'.

contracts/native-staking/src/contract.rs
ethanfrey: The `native-staking` contract is a look up between owners and proxy contracts (and visa versa).
It will need to store that state, I will add queries for that

contracts/native-staking/src/contract.rs
ethanfrey: StargateQueries don't work for SDK chains either. They have been disabled for over a year since someone found non-determinism in them and halted Juno.

I will leave it as hardcoded for now, and make an issue for v1 to revisit this

contracts/vault/src/error.rs
ethanfrey: Mauro is correct. Lien is what I intended:

"a right to keep possession of property belonging to another person until a debt owed by that person is discharged"

contracts/native-staking-proxy/src/contract.rs
ethanfrey: Yes, typo. And good catch (why we need tests and real implementation). I just started sketching it out

PR #18, opened 2023-05-02
Add basic CI

### LINKED ISSUE #15: Set up basic CI tests
These should cover pure rust tests (and linting, etc)

JakeHartnell:
Closes #15

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -1,0 +1,67 @@
+# Based on https://github.com/actions-rs/example/blob/master/.github/workflows/quickstart.yml
+
+on: [push, pull_request]
+
+name: Basic Checks
+
+jobs:
+  test:
+    name: Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install latest nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          target: wasm32-unknown-unknown
+          override: true
+
+      - name: Run tests
+        uses: actions-rs/cargo@v1
+        with:
+          toolchain: nightly
+          command: test
+          args: --workspace
+        env:
+          RUST_BACKTRACE: 1
+
+      - name: Compile WASM contract
+        uses: actions-rs/cargo@v1
+        with:
+          toolchain: nightly
+          command: wasm
+        env:
+          RUSTFLAGS: "-C link-arg=-s"
+
+  lints:
+    name: Lints
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+          components: rustfmt, clippy
+
+      - name: Run cargo fmt
+        uses: actions-rs/cargo@v1
+        with:
+          toolchain: nightly
+          command: fmt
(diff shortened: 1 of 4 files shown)


maurolacy (approved):
LGTM.

.github/workflows/basic.yml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+  test:
+    name: Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install latest nightly toolchain
```
ethanfrey, line 15: Let's not use nightly.
Code should work with some well-defined version of Rust and only use nightly briefly if we really need some feature that is about to be stabilized.

I think MSRV for cosmwasm is 1.65. Latest release is 1.70.

We can use anything between that (inclusive) for this CI. But make it explicit and then we make a PR to update.

Sometimes newer version flag lint errors earlier ones didn't and randomly breaking CI (when a new rust docker image is pushed) is not so much fun.

.github/workflows/basic.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          profile: minimal
+          toolchain: nightly
+          target: wasm32-unknown-unknown
+          override: true
+
+      - name: Run tests
+        uses: actions-rs/cargo@v1
+        with:
+          toolchain: nightly
```
ethanfrey, line 26: I would remove `toolchain: nightly` everywhere and put whichever version we need. I think just `stable`

.github/workflows/basic.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Install latest nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          target: wasm32-unknown-unknown
+          override: true
+
+      - name: Run tests
```
ethanfrey, line 23: Nit, I often include a build phase before test.

If there are build errors and I run test, sometimes they seem to be reported twice.
I mean, it will fail regardless, but seems to give better output running them one after another

.github/workflows/basic.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions-rs/cargo@v1
+        with:
+          toolchain: nightly
+          command: test
+          args: --workspace
+        env:
+          RUST_BACKTRACE: 1
+
+      - name: Compile WASM contract
```
ethanfrey, line 32: Nit: this compiles whole workspace and not really good at checking contracts.

You could try something like https://github.com/CosmWasm/cw-plus/blob/main/.circleci/config.yml#L500-L519

If something gets imported with `library` feature flag, then it will show up as super small contract and missing entry points (where cosmwasm-check will fail). Nicer to have this fail on the PR that changes the imports, then right before a release.

(You can make those additions as a future PR)

.github/workflows/basic.yml
```diff
@@ -0,0 +1,67 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
```
ethanfrey, line 51: How about stable (to match the name).
Ideally use a fixed tag, so CI doesn't break when new stable comes out

ethanfrey (commented):
Can you please not use nightly in CI.
Let's use a fixed stable version, like we will use for really developing and compiling them.

ethanfrey (approved):
Great changes.

Question for future: can we put 1.65.0 in some variable for reuse?

PR #19, opened 2023-05-03
Finalize producer apis

### LINKED ISSUE #2: Finalize documentation for provider side

ethanfrey:
Closes #2 

- [x] Addresses comments from #17
- [x] Adds better docs on vault functionality

### DIFF
--- a/contracts/vault/README.md
+++ b/contracts/vault/README.md
@@ -3,3 +3,62 @@
 This is the central entrypoint to Mesh Security on the Provider side. Users can provide collateral to this contract,
 which then allows you to use that collateral to both stake locally (see [mesh-native-staking](../native-staking)),
 as well as cross-stake.
+
+## Definitions
+
+Vault Denom - The token we use as collateral. Generally the native staking token (but there might be
+cases for IBC assets, eg. ETH, BTC, or USDC). 
+
+Collateral - Total amount of tokens deposited by an account.
+
+Local Staking Contract - One contract per vault that can receive the collateral to be staked
+on the native staking system. The tokens moved here can only be staked and count towards the
+account's collateral.
+
+Cross Staking Contract - Any number of contracts per vault that receive Liens on the collateral.
+
+Lien - A claim on some collateral (from local- or cross-staking contract). Collateral with existing
+liens may not be withdrawn.
+
+## Workflow
+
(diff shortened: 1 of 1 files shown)


JakeHartnell (approved):
LGTM! Having `owner_by_proxy` query is a good add.

contracts/vault/README.md
```diff
@@ -3,3 +3,62 @@
...
+Collateral - Total amount of tokens deposited by an account.
+
+Local Staking Contract - One contract per vault that can receive the collateral to be staked
+on the native staking system. The tokens moved here can only be staked and count towards the
+account's collateral.
+
+Cross Staking Contract - Any number of contracts per vault that receive Liens on the collateral.
+
+Lien - A claim on some collateral (from local- or cross-staking contract). Collateral with existing
+liens may not be withdrawn.
```
ValarDragon, line 21: Do Liens represent both:
* how much could be under slashing risk
* duration it must be locked for?

Or is the duration it must be locked for each 'lien' tracked separately?

contracts/vault/README.md
```diff
@@ -3,3 +3,62 @@
...
+There are tokens in the system to match the collateral (accounting amount). 
+`sum(collateral) <= balance(vault) + balance(local staking)`, where balance includes delegations.
+This implies a slashing event on local staking must reduce collateral.
+
+For each user, they have collateral equal to or greater than every lien.
+`max(liens(user)) <= collateral(user)`
+
+For each user, the total amount of potential slashing over all liens is less than or
+equal to their total collateral (important if doing many cross-stakes, or with high slashing rates):
+`liens(user).map(|x| x.lien * x.max_slashing_rate).sum() <= collateral(user)`
```
ValarDragon, line 58: We should add a todo for ourselves, to reason about how to express chain's restrictions on when you could go above max slashing rate. (Issue being that correlated slashing can make the max slashing rate quite high)

ValarDragon (approved):

contracts/vault/README.md
ethanfrey: The duration is tracked by the LocalStaking / CrossStaking contract.

The vault doesn't care if it is bonded or unbonding. Only when it is finally released do we remove the lien.

(It is only the first point)

PR #20, opened 2023-05-04
Add producer mocks

### LINKED ISSUE #5: Produce mock contracts for all Provider APIs
This means:
* Vault
* Local Staking
* Remote Staking

They should have dummy implementation for multi-test testing.

Local staking and Remote staking should be designed so we can start using them in UI (have enough queries implemented, reasonable messages)

ethanfrey:
Closes #5

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ repository    = "https://github.com/osmosis-labs/mesh-security"
 
 [workspace.dependencies]
 mesh-apis        = { path = "./packages/apis" }
+mesh-mocks        = { path = "./packages/mocks" }
 mesh-vault    = { path = "./contracts/vault" }
 mesh-native-staking    = { path = "./contracts/native-staking" }
 mesh-native-staking-proxy    = { path = "./contracts/native-staking-proxy" }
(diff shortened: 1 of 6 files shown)


ethanfrey:
@hashedone I don't know if I need more to make these mocks usable in multitest (like export `instantiate`, `execute`, etc).
Or if it is only blocked on stabilized mt feature on Sylvia.

hashedone (approved):
In general, I am hesitating a bit on that. Today/tomorrow, we are releasing Sylvia generating similar mocks, but which are a little bit differently structured - we don't keep anything but an address in them, and we do builder-pattern for messages. We could just use them.

On the other hand - we want to do a little bit of cleanup on that (https://github.com/CosmWasm/sylvia/issues/109, https://github.com/CosmWasm/sylvia/issues/114) - those two are small changes, but breaking. Therefore - I would accept this, for now, use generated Sylvia helpers for any new tests, and unify everything after 0.5 release which I hope would be stable for a bit longer.

ethanfrey:
Yes, I'm very happy if you can replace this by auto-gen later.
But until that time, if these work, let's use them.

So the work on Sylvia can be a background task and not blocking mesh-security (maybe Jan can just keep running ahead on these key issues as we keep working with manual generation until it is ready). Actually, I think using these manual helpers as examples for the codegen is the only way to get good codegen. Design it in the concrete, them automate

PR #21, opened 2023-05-04
Add Sylvia helpers

ethanfrey:
Add some helper classes to call `#[interface]` traits (struct that wraps execute and query methods to use in cross-contract calls).

This is written by hand now to simplify the coding in the contracts (which will be repeated in mocks and real contracts). But ideally this could be autogenerated in a future sylvia release.

I added `funds: Vec<Coin>` as an extra arg to all execute methods (even ones that don't need funds) to simulate the API that would be generated by sylvia codegen. But it is already usable as is.

### DIFF
--- a/packages/apis/src/cross_staking_api.rs
+++ b/packages/apis/src/cross_staking_api.rs
@@ -1,5 +1,3 @@
-use std::ops::Deref;
-
 use cosmwasm_schema::cw_serde;
 use cosmwasm_std::{to_binary, Addr, Binary, Coin, Deps, Response, StdError, Uint128, WasmMsg};
 use sylvia::types::{ExecCtx, QueryCtx};
@@ -36,14 +34,6 @@ pub trait CrossStakingApi {
 
 #[cw_serde]
 pub struct CrossStakingApiHelper(pub Addr);
-
-impl Deref for CrossStakingApiHelper {
-    type Target = Addr;
-
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
 
 impl CrossStakingApiHelper {
     pub fn addr(&self) -> &Addr {
--- a/packages/apis/src/cross_staking_api.rs
+++ b/packages/apis/src/cross_staking_api.rs
@@ -1,4 +1,7 @@
-use cosmwasm_std::{Binary, Response, StdError, Uint128};
+use std::ops::Deref;
+
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{to_binary, Addr, Binary, Coin, Deps, Response, StdError, Uint128, WasmMsg};
 use sylvia::types::{ExecCtx, QueryCtx};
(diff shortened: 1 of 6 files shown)


JakeHartnell:
Does this supersede #20?

hashedone (approved):
The same comment as #20, except I see more helpers which are not yet generated in Sylvia so it would make sense for longer. Queriers are ongoing, and will probably be released next week; generating exec messages is next in the queue.

I will accept that.

Usage of generated queries would be a touch different - instead of passing `deps` to the calls, the temporary querier object would be generated (with the `querier` function on generated helper object), so it doesn't need to be passed with every single query (but there still be a type to be stored as remote contract address).

However this is good as it is, and will allow progress, and when we release queries I will just do alignment of all usages quickly.

packages/apis/src/cross_staking_api.rs
```diff
@@ -30,3 +33,41 @@ pub trait CrossStakingApi {
...
+#[cw_serde]
+pub struct CrossStakingApiHelper(pub Addr);
+
+impl Deref for CrossStakingApiHelper {
+    type Target = Addr;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
hashedone, line 46: I would strongly avoid implementing `Deref` trait like this - long discussion why, but long story short: even Rust documentation claims:

> On the other hand, the rules regarding Deref and [DerefMut](https://doc.rust-lang.org/std/ops/trait.DerefMut.html) were designed specifically to accommodate smart pointers. Because of this, Deref should only be implemented for smart pointers to avoid confusion.

Instead, I would implement `AsRef<Addr>` on this type, and possibly directly the `pub fn addr(&self) -> &Addr`  function.

packages/apis/src/cross_staking_api.rs
```diff
@@ -30,3 +33,41 @@ pub trait CrossStakingApi {
...
+    pub fn addr(&self) -> &Addr {
+        &self.0
+    }
+
+    pub fn receive_virtual_stake(
+        &self,
+        owner: String,
+        amount: Uint128,
+        msg: Binary,
+        funds: Vec<Coin>,
```
hashedone, line 58: You are mixing message-related arguments with "metadata" (here: funds). Some messages would need those funds with them, some not, but what is a bit confusing - it is not clear if funds are funds sent with message or the argument of the message.

Here it is, I think, ok, but to make it clear in Sylvia, we would go with more like a builder pattern, so message arguments and their metadata would be passed separately.

packages/apis/src/cross_staking_api.rs
ethanfrey: Okay. I just used `Deref` for things like `struct Addr(String)`, where it let's us use all the wrapped type's functions without manually passing them through. Fair enough to remove it and force this explicit.

And happy if you give me some Rust coaching on these language semantics.

Issue #22, opened 2023-05-15
Always use denom with currency amounts


alpe:
I have noticed that Bond/Unbond msgs in the [virtual staking doc](https://github.com/osmosis-labs/mesh-security/blob/main/docs/consumer/VirtualStaking.md#interface) do not contain a denom value for example. There should only be one staking token but it would be better have this explicitly set by the contract via `Coin` type. This makes it more readable for humans and future proof.

PR #23, opened 2023-05-16
Initial go module draft

alpe:
Start design work for Go module

### DIFF
--- a/docs/consumer/go_module.md
+++ b/docs/consumer/go_module.md
@@ -3,18 +3,16 @@
 ## Consumer - Virtual Staking Contract
 
 ### Bootstrapping the contract(s)
-Bootstrapping of the mesh consumer side contracts requires some orchestration and configuration work that can be done by the Go module
-as an alternative to a manual setup. This comes also with some benefits for system- / integration-tests. 
-The challenge is to keep the contracts that are shipped for bootstrapping up to date with the version of their 
-own repository.
+Contracts will be uploaded and instantiated via gov proposals on permissioned chains. The proposal(s) will be submitted manual.
+Wasmd contains the required proposal types already. It may make sense to use gov v1 multi-message proposals. 
 
 ### Contract Authorization
 In order to let a contract manage virtual stake, an authorization by the consumer chain must be given to ensure security.
-This can be done either by a the Go module on chain update with migration code where the source and integrity of the contract is 
-ensured before or via proposal and gov authority on a running chain.
+This is usually done by a proposal and gov authority for a running chain.
+As authorization rules can be very chain specific, this is going to be an extension point in the implementation with an interface.
 
 For simplicity the contract authorization for virtual stake can be combined with the **max cap limit** configuration per contract.
-No max cap limit means no authorization.
+This would be the default implementation. No max cap limit means no authorization.
 
 Example for gov v1 message to add/ update the max cap limit for a contract:
 
@@ -33,10 +31,8 @@ message MsgSetVirtualStakingMaxCap {
 }
 ```
 
-### Bond Virtual Stake
-
-#### Contract messages
-Example of the custom messages that can be sent by an authorized contract to mint virtual tokens and delegate to a validator:   
+### Contract messages
+Example of the custom messages that can be sent by an authorized contract to mint/burn virtual tokens and un-/ delegate to a validator:   
 ```go
 type CustomMsg struct {
 	VirtualStake *VirtualStakeMsg `json:"virtual_stake,omitempty"`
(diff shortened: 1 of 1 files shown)


docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
+## Go Module designs
+
+## Consumer - Virtual Staking Contract
+
+### Bootstrapping the contract(s)
```
ethanfrey, line 5: I think we don't need them baked into the chain.

They have to be uploaded on the provider and I think this is most flexible.
Don't worry about chain init or anything special on import/migration.

Let's just upload them like normal contracts in integration tests (as this will be real-world scenario)

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+Bootstrapping of the mesh consumer side contracts requires some orchestration and configuration work that can be done by the Go module
+as an alternative to a manual setup. This comes also with some benefits for system- / integration-tests. 
+The challenge is to keep the contracts that are shipped for bootstrapping up to date with the version of their 
+own repository.
+
+### Contract Authorization
+In order to let a contract manage virtual stake, an authorization by the consumer chain must be given to ensure security.
+This can be done either by a the Go module on chain update with migration code where the source and integrity of the contract is
```
ethanfrey, line 13: Let's just focus on "gov authority on a running chain". And avoid migration work

ethanfrey: We need to have existing contracts instantiated with an active ibc channel to the proper provider, and then have governance manually verify this is the correct setup before providing max cap.

Let's just focus on the "gov vote" case, as I don't see any other use outside of CI.

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+  // Authority is the address that controls the module (defaults to x/gov unless
+  // overwritten).
+  string authority = 1;
+
+  // Contract is the address of the smart contract that is given permission
+  // do virtual staking which includes minting and burning staking tokens.
+  string contract = 2;
+
+  // MaxCap is the limit up this the virtual tokens can be minted.
+  cosmos.base.v1beta1.Coin max_cap = 5;
```
ethanfrey, line 32: Thought:

My first instinct was "Let's just use `sdk.Int` here."
As we would only mint the native staking token.

But if you see another use case to mint other virtual tokens, let me know.
If there is a valid case, happy to keep it general. But making things general for cases that may never exist is a large cause of complexity creep in the cosmos-sdk.

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+#### Contract messages
+Example of the custom messages that can be sent by an authorized contract to mint virtual tokens and delegate to a validator:   
+```go
+type CustomMsg struct {
+	VirtualStake *VirtualStakeMsg `json:"virtual_stake,omitempty"`
+}
+
+type VirtualStakeMsg struct {
+	Bond   *BondMsg   `json:"bond,omitempty"`
+	Unbond *UnbondMsg `json:"unbond,omitempty"`
```
ethanfrey, line 47: Can you add Unbond struct as well?

I assumes it does instant unbond and then burns those tokens?

Looks good, just please add those to the docs.

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+}
+
+type BondMsg struct {
+	Amount    wasmvmtypes.Coin `json:"amount"`
```
ethanfrey, line 51: Again, this only works on native staking token. We could make it Uint128 / sdk.Int, right?

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+	Validator string           `json:"validator"`
+}
+```
+
+#### Custom Message Handler
```
ethanfrey, line 56: This can gladly go into https://github.com/osmosis-labs/mesh-security-sdk repo.

I think the high level function and API belongs here (everything above this line, probably more). But the implementation code is not needed here. I will look at https://github.com/osmosis-labs/mesh-security-sdk/pull/1 as well

docs/consumer/go_module.md
ethanfrey: Nevermind this is handled in line 126.
If you delete the sample code (leave that in other repo), I think the API will read clear and no changes needed here

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+#### Contract messages
+Example of the custom messages that can be sent by an authorized contract to unbond from a delegator and burn virtual tokens:
+```go
+type VirtualStakeMsg struct {
+	Bond   *BondMsg   `json:"bond,omitempty"`
+	Unbond *UnbondMsg `json:"unbond,omitempty"`
+}
+
+type UnbondMsg struct {
+  Amount    wasmvmtypes.Coin `json:"amount"`
```
ethanfrey, line 134: Same comment about Coin vs Uint128/sdk.Int

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+  Validator string           `json:"validator"`
+}
+
+```
+
+#### Custom Message Handler
```
ethanfrey, line 140: Same comment about removing this code (and just committing in the Go repo)

docs/consumer/go_module.md
```diff
@@ -0,0 +1,205 @@
...
+	k.bank.AddSupplyOffset(cacheCtx, bondDenom, unbondedAmount)
+	newDelegatedAmt := k.getTotalDelegatedAmount(cacheCtx, actor).Sub(unbondedAmount)
+	k.setTotalDelegatedAmount(cacheCtx, actor, newDelegatedAmt)
+
+	done()
+	return nil
+}
+```
+
+## Integration of Cosmos-SDK and Osmosis Fork
```
ethanfrey, line 189: Very good thoughts here. I would expand a bit more on this.

Do we need to fork `x/bank` and `x/staking`?

I understand the `/x/virtual_stake` package that accepts some adapter with these 3 methods (and some standard Bank/Staking functions). This adapter would then call into either osmosis-specific code, or into... what? A forked bank and staking module? Some standard wrapper around Bank and Staking that could be added?

This is a great idea to make this so flexible and not require changes on osmosis or others to adopt many osmosis-specific modules or fork staking without need. I'd love to see a few more paragraphs explaining this design

ethanfrey (commented):
Nice start. I will now look at https://github.com/osmosis-labs/mesh-security-sdk/pull/1

You can trim some code details here (that are in the other repo), adjust a few API definitions, and expound on the extension points (great idea!). With those, ready to merge 👍

docs/consumer/go_module.md
alpe: Fair enough. Better start simple.
I was assuming a dependency between contracts that may require authoritzation for some operations. We used `PreAuthsHooks` in poe-contracts but they require an order and TX boundary not not let anybody slip in.
With gov v1 multi-message proposals, this would be possible now but there is no CLI support.

docs/consumer/go_module.md
alpe: I have a strong opinion on this to use a Coin type. Technically, you are right that the amount would be enough. Unfortunately, I have been in enough situation and discussions where it was confusing without the unit. We already have osmo, uosmo, (virtual-,) synthetic osmo as terms. Who knows what comes in the future....
My rule of thumb is to make the unit explicit so that there is no need for assumptions.

Using the Coin type is consistent with existing staking messages defined in [wasmvm](https://github.com/CosmWasm/wasmvm/blob/v1.2.3/types/msg.go#L232-L245)

I had opened #22 to update the docs

docs/consumer/go_module.md
alpe: There are some prototypes in the [spike](https://github.com/osmosis-labs/mesh-security-sdk/pull/1/files#diff-f75c178ffdca207e89258109bab5fc6acce01314f2ba7bd653606dbff1a8503e) but this is still early stage.

docs/consumer/go_module.md
ethanfrey: We discussed this and I agree with using the coin type for clarity

ethanfrey:
In absence of comments from others, merging this so I can rebase my docs on top

PR #26, opened 2023-05-18
Use denom for amounts

### LINKED ISSUE #22: Always use denom with currency amounts

ethanfrey:
Closes #22 

Update both the docs as well as the current code stubs.

Also rename remaining "lein" -> "lien"

### DIFF
--- a/packages/mocks/src/local_staking.rs
+++ b/packages/mocks/src/local_staking.rs
@@ -3,7 +3,7 @@ use sylvia::{contract, schemars};
 use thiserror::Error;
 
 use cosmwasm_schema::cw_serde;
-use cosmwasm_std::{ensure_eq, Binary, Coin, Decimal, Response, StdError, Uint128};
+use cosmwasm_std::{ensure_eq, Binary, Coin, Decimal, Response, StdError};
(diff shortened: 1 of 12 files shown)


packages/mocks/src/local_staking.rs
```diff
@@ -64,18 +67,21 @@ impl MockLocalStakingContract<'_> {
...
         let cfg = self.config.load(ctx.deps.storage)?;
-        let funds = Coin {
-            denom: cfg.denom,
-            amount,
-        };
+        ensure_eq!(
+            cfg.stake_denom,
+            amount.denom,
+            StakingError::WrongDenom(cfg.stake_denom)
+        );
```
maurolacy, line 79: 👍🏼much safer.

maurolacy (approved):
LGTM.

Issue #7, opened 2023-04-24
Finalize documentation


ethanfrey:
* Define all components for the consumer side (including APIs / Stubs)
* Clearly define IBC packets types (Rust structs/enums)
* Document and define Custom SDK module to be integrated

PR #28, opened 2023-05-18
Improve documentation on the consumer side

ethanfrey:
Addresses #7 

Also cover common questions and failure modes on the top-level README.

I would appreciate [many eyes on the common questions](https://github.com/osmosis-labs/mesh-security/blob/extend-consumer-docs/docs/README.md#common-concerns) to see if they are answered in sufficient detail and if the design is sufficiently robust.


docs/README.md
```diff
@@ -55,6 +55,125 @@ chains can only provide security in their native token(s).
...
+make a governance vote to allow immediate unbonding of all cross-stake to that consumer.
+
+The last point is a bit trickier. We will not fully define slashing until v1, but the design is
+that the Consumer cannot initiate slashing via an IBC message, rather they need to submit the
+original evidence (multiple headers on the same height signed by the same validator). This will
+be verified and punished on the Provider side. A normal state machine bug on the consumer could not
+trigger this. An actual double sign (the validator was malicious or made mistake) could trigger it.
+
+However, some malicious code in Tendermint, to double-sign or leak the private key, could trigger
+such a condition without the Validators fault. It would require hacks deep into the Tendermint code
```
jtremback, line 173: Just to clarify this point a bit- It is impossible to defend against a compromised Tendermint. A compromised Tendermint could simply make the validator actually double sign. Then there would be real, legitimate evidence of the double sign.

The end game for security against malicious consumer chains is for validators to verify that they are running the right version of Tendermint, and not have to trust the chain's application

docs/consumer/Converter.md
```diff
@@ -54,6 +54,11 @@ Thus this cross-stake will trigger the converter to request the virtual staking
 
 The discount is stored in the Converter contract and can only be updated by the admin (on-chain governance).
 
+**Important** When we calculate the virtual stake (eg 1080 CONS in the example above), those
+tokens will be staked as if they were native CONS tokens. They have the same influence on the
+validator's voting power, and will receive the same rewards. The only difference is that they
+can never be withdrawn and slashing is managed remotely on the Provider chain.
```
maurolacy, line 60: Should a remote slashing event trigger a re-evaluation? I mean, different collateral amount so, burning / unstacking required?

docs/consumer/Converter.md
```diff
@@ -86,37 +91,31 @@ to the particular chain it is being deployed on. With this TWAP we convert eg. 1
...
+(If we get callbacks on ics20, we send the metadata only after tokens have arrived. Until then 
+(for MVP), we send them concurrently and hope)
 
 ## Unstaking Flow
 
-The Converter can also unstake some tokens. These will be held in escrow on the Provider and are susceptible to slashing upon proper evidence
-submission. Since the virtual stake is, well, "virtual" and slashing has no impact, the delegation numbers can be immediately reduced
+The Converter can also unstake some tokens. These will be held in escrow on the Provider and
+are susceptible to slashing upon proper evidence submission. Since the virtual stake is, well, 
+"virtual" and slashing has no impact, the delegation numbers can be immediately reduced
```
maurolacy, line 120: Again, shouldn't slashing cause "virtual" token burn?

docs/consumer/VirtualStaking.md
```diff
@@ -166,17 +166,47 @@ I request to bond 120 more to B, which turns into 20 bond to B. There is an actu
...
-solution may use some batches). Limiting the interface to custom SDK modules, we would first "immediately unbond"
-all other validators, then bond that virtual stake to the newly bonded validator. Much care must be take with rounding issues
-to avoid going 0.000001 token above the max cap.
+To properly handle this, we should use Redelegations to rebalance when the amounts are updated. 
+This may be quite some messages (if we have 50 validators, each time we stake one more, we need to 
+add a bit to that one and decrease the other 50). We describe a solution to this in the next section.
+
+### Epochs
+
+For efficiency rewards will be withdrawn for all cross-stakers at a regular rhythm, once per epoch.
```
maurolacy, line 175:
```suggestion
For efficiency, rewards will be withdrawn for all cross-stakers at a regular rhythm, once per epoch.
```

docs/consumer/VirtualStaking.md
```diff
@@ -166,17 +166,47 @@ I request to bond 120 more to B, which turns into 20 bond to B. There is an actu
...
+This is done to reduce the number of IBC messages sent, and reduce computation (gas) overhead.
+Note that one provider will like have to withdraw for dozens of validators and then distribute
```
maurolacy, line 177:
```suggestion
Note that one provider will likely have to withdraw from dozens of validators and then distribute
```
?

docs/consumer/VirtualStaking.md
```diff
@@ -166,17 +166,47 @@ I request to bond 120 more to B, which turns into 20 bond to B. There is an actu
...
+to the proper cross-stakers, so this is a significant amount of computation, and shouldn't be
+triggered every few blocks.
+
+While staking and unstaking can be quite cheap below the max cap, it becomes quite expensive once
+the number of tokens exceeds the max cap, as it will require a rebalance. If a provider is
+over the max cap and someone stakes 10 tokens on validator V, the provider will have to unstake
+some tokens proportionally from every other validator. Likewise, if someone unstakes 10 tokens,
+it will trigger a staking on the remaining validators to keep the total at max cap.
+
+It makes sense to do this rebalance only once per epoch, as it is quite expensive. These could
```
maurolacy, line 187: Add a "CosmWasm Contract" sub-title here?

docs/consumer/VirtualStaking.md
```diff
@@ -166,17 +166,47 @@ I request to bond 120 more to B, which turns into 20 bond to B. There is an actu
...
+be different epoch periods for two different purposes, but it is recommended to keep them
+the same for simplicity. However, we do suggest that the epochs of each provider be offset,
+so we don't blog up the network with a bunch of IBC messages at the same time.
```
maurolacy, line 190:
```suggestion
so we don't clog up the network with a bunch of IBC messages at the same time.
```

docs/consumer/VirtualStaking.md
```diff
@@ -166,17 +166,47 @@ I request to bond 120 more to B, which turns into 20 bond to B. There is an actu
 stakers have delegated at a ration of 2:3. This gets worse if eg. someone unbonds all A. We end up with 120 requested for B,
```
maurolacy, line 166:
```suggestion
stakers have delegated at a ratio of 2:3. This gets worse if eg. someone unbonds all A. We end up with 120 requested for B,
```

docs/consumer/VirtualStaking.md
```diff
@@ -166,17 +166,47 @@ I request to bond 120 more to B, which turns into 20 bond to B. There is an actu
...
-Note: this example demonstrates why the immediate unbond power is required to properly implement this, and we cannot just do a pure contract
-calling normal staking commands (or we would be limited to shifting validators once per unbond period).
+```rust
+#[cw_serde]
+pub enum SudoMsg {
+  Rebalance{},
+}
+```
 
 ### Module
```
maurolacy, line 211:
```suggestion
### SDK Module
```

docs/consumer/VirtualStaking.md
```diff
@@ -194,33 +224,37 @@ The Virtual Staking **Module** also maintains the current state of each of the r
 
 ```go
 type StakePermission struct {
-  /// Limit we cap the virtual stake of this converter.
-  /// Defined as a number of "virtual native tokens" this can mint.
+  // Limit we cap the virtual stake of this converter.
+  // Defined as a number of "virtual native tokens" this can mint.
```
maurolacy, line 228:
```suggestion
  // Limits the cap of the virtual stake of this converter.
  // Defined as a number of "virtual native tokens" this module can mint.
```
?

docs/consumer/VirtualStaking.md
```diff
@@ -194,33 +224,37 @@ The Virtual Staking **Module** also maintains the current state of each of the r
...
 ```
 
 Beyond MVP, we wish to add the following functionality:
 
-* Provide WithdrawReward callbacks each epoch (eg 1 day) on BeginBlock to all registered contracts
-* Provide configuration for optional governance multiplier (eg 1 virtual stake leads to 1 tendermint power,
-but may be 0 or 1 or even 0.5 gov voting power)
+* Provide configuration for optional governance multiplier (eg 1 virtual stake leads to 1 tendermint power, but may be 0 or 1 or even 0.5 gov voting power)
 
 ### Reward Withdrawls
```
maurolacy, line 240:
```suggestion
### Reward Withdrawals
```

docs/consumer/VirtualStaking.md
```diff
@@ -194,33 +224,37 @@ The Virtual Staking **Module** also maintains the current state of each of the r
...
+The Virtual Staking Module uses an [EndBlock hook called once per epoch](#epochs)
+(param setting, eg 1 day).
+This will trigger a `SudoMsg` to each Virtual Staking contract, allowing them to both do a rebalancing
+with any staking/unstaking in that period, as well as withdraw rewards from all validators.
 
-The Virtual Staking Module also has a BeginBlock hook called once per epoch (param setting, eg 1 day), that will trigger all reward withdrawals.
-This epoch has a different start for each converter (based on when they were authorized), so they happen staggered over time.
-When the epoch finishes for one Converter, the Module will withdraw rewards from all delegations that converter made, and send those tokens
-to the Converter along with the info of which validator these are for.
+When the epoch finishes for one Converter, the Module will withdraw rewards from all delegations
```
maurolacy, line 247:
```suggestion
When the epoch finishes for one Converter, the Virtual Staking Module will withdraw rewards from all delegations
```

maurolacy (approved):
Just some questions and syntax suggestions.

There's a mention of a 'Meta-Staking' module that I haven't seen mentioned elsewhere, can you clarify / add references?

How the "virtual tokens" work? In the sense that they "don't affect max supply, and can only be used for staking". Is this a SDK provided kind of token? Or it'll be implemented by a modified staking module. Just curious.

docs/README.md
```diff
@@ -55,6 +55,125 @@ chains can only provide security in their native token(s).
...
+### Power Limits on Remote Chains
+
+Another common concern is whether there are effective limits in the power a remote
+chain can exert over one chain. Would it be possible for a higher cap chain (eg $ATOM)
+to take over more than one-third, or even two-thirds or the power of a smaller cap chain
+(eg $STARS) it is cross-staking on. 
+
+Clearly the consumer chain wants to put some limits. The first limit is the 
+[discount applied during the conversion](https://github.com/osmosis-labs/mesh-security/blob/main/docs/consumer/Converter.md#price-normalization).
+This doesn't just provide margin for price fluxuations but also means that on average
```
maurolacy, line 83:
```suggestion
This doesn't just provide margin for price fluctuations but also means that on average
```

docs/README.md
```diff
@@ -55,6 +55,125 @@ chains can only provide security in their native token(s).
...
+disproportionate power over the affairs of the consumer. If the provier price falls too much,
+in the period until the price oracle is updated, there may be less remote collateral
+than the virtual stake.
+
+For price increases, we have the max cap described under "Power Limits" which places a cap
+on the provider chain's influence in all circumstances, so this isn't a large problem.
+
+For a rapid price decrease, we must consider the time frame.
+It is a [requirement of an oracle](https://github.com/osmosis-labs/mesh-security/blob/main/docs/consumer/Converter.md#price-feeds)
+to provide timely feeds, say once a day or week, so we should focus the relative price movement
```
maurolacy, line 117:
```suggestion
to provide timely feeds, say once a day or week, so we should focus on the relative price movement
```

maurolacy (commented):
Couple more syntax suggestions.

docs/consumer/Converter.md
ethanfrey: It should. But let's specify this better.

Actually, the whole slashing portions is a WIP for the second milestone, but I need to define it all better.

I wanted to finalize all docs that were blocking people from coding the MVP before going into that rabbit hole. But good questions, we should loop back in a week or two.

docs/README.md
ethanfrey: Yeah, it would be nice to run them as two different processes.
Maybe different user permissions or in docker to isolate tendermint and abci app better.

Same issue for mesh and replicated security

ethanfrey: > It is impossible to defend against a compromised Tendermint.

Not fully true. I think tmkms from iqlusion enforces no double signing on the firmware level. Like the external hardware device will never sign a second block for a height / round it has previously signed.

Not 100% sure, but I think their model was safety if someone got root on their server

PR #30, opened 2023-05-23
Vault

### LINKED ISSUE #6: Write the vault contract
* Implement all functionality [defined in the docs](https://github.com/osmosis-labs/mesh-security/blob/main/docs/provider/Vault.md).
* Much inspiration (and code?) can be borrowed from the [hackathon lockup implementation](https://github.com/CosmWasm/mesh-security-hackathon/tree/main/contracts/mesh-lockup)
* Add tests using the mocks written in #5

hashedone:
Closes #6


contracts/vault/src/contract.rs
```diff
@@ -63,56 +77,165 @@ impl VaultContract<'_> {
     }
 
     #[msg(exec)]
-    fn bond(&self, _ctx: ExecCtx) -> Result<Response, ContractError> {
-        // this assets the proper token is sent in info.funds
-        todo!()
+    fn bond(&self, ctx: ExecCtx) -> Result<Response, ContractError> {
```
ethanfrey, line 80: looks good.

contracts/vault/src/contract.rs
```diff
@@ -63,56 +77,165 @@ impl VaultContract<'_> {
...
+            .add_attribute("amount", amount.to_string());
+
+        Ok(resp)
     }
 
     #[msg(exec)]
-    fn unbond(&self, _ctx: ExecCtx, amount: Coin) -> Result<Response, ContractError> {
-        let _ = amount;
-        todo!()
+    fn unbond(&self, ctx: ExecCtx, amount: Uint128) -> Result<Response, ContractError> {
```
ethanfrey, line 100: good

contracts/vault/src/contract.rs
```diff
@@ -63,56 +77,165 @@ impl VaultContract<'_> {
...
     #[msg(exec)]
     fn stake_remote(
         &self,
-        _ctx: ExecCtx,
+        mut ctx: ExecCtx,
         // address of the contract to virtually stake on
         contract: String,
         // amount to stake on that contract
-        amount: Coin,
+        amount: Uint128,
```
ethanfrey, line 133: Discussed this with Alex, and while the coin denom is redundant (it is defined on the contract level), we felt it does provide clarity of intention. Especially if there are multiple vaults on one chain, which is possible in discussion of cross-staking dao tokens.

https://github.com/osmosis-labs/mesh-security/pull/23#discussion_r1195550632

Let's just leave as Coin please. We can have another discussion with this issue and get a few more voices if you feel strongly.

ethanfrey: Bascially, it would just be `ensure_eq!(config.denom, amount.denom, ContractError::WrongDenom);`

contracts/vault/src/contract.rs
```diff
@@ -63,56 +77,165 @@ impl VaultContract<'_> {
...
     ) -> Result<Response, ContractError> {
-        let _ = (contract, amount, msg);
-        todo!()
+        nonpayable(&ctx.info)?;
+
+        let contract = ctx.deps.api.addr_validate(&contract)?;
+        let contract = CrossStakingApiHelper(contract);
+        let slashable = contract.max_slash(ctx.deps.as_ref())?;
+
+        self.stake(&mut ctx, &contract.0, slashable.max_slash, amount)?;
```
ethanfrey, line 143: Maybe we should add a better API to get this, rather than `&contract.0`. Like `contract.as_str()`

(Note for adding to Sylvia, not for this PR)

contracts/vault/src/contract.rs
```diff
@@ -123,19 +246,90 @@ impl VaultContract<'_> {
...
+        self.local_staking.save(deps.storage, &local_staking)?;
 
         Ok(Response::new())
     }
+
+    /// Updates the local stake for staking on any contract
+    fn stake(
+        &self,
+        ctx: &mut ExecCtx,
+        addr: &Addr,
```
ethanfrey, line 269: `s/addr/lienholder/` or `s/addr/staked_on/` for clarity.
Naming a variable by type provides no extra context

contracts/vault/src/contract.rs
```diff
@@ -123,19 +246,90 @@ impl VaultContract<'_> {
...
+                amount: Uint128::zero(),
+                slashable,
+            });
+        lien.amount += amount;
+
+        let mut user = self
+            .users
+            .may_load(ctx.deps.storage, &ctx.info.sender)?
+            .unwrap_or_default();
+        user.max_lien = user.max_lien.max(lien.amount);
```
ethanfrey, line 286: Optional suggestion: move these lines into a method on UserInfo, as you have such calculation.
Then people don't modify the raw fields, but rather use these methods to do so.

contracts/vault/src/contract.rs
```diff
@@ -123,19 +246,90 @@ impl VaultContract<'_> {
...
+        self.liens
+            .save(ctx.deps.storage, (&ctx.info.sender, addr), &lien)?;
+
+        self.users.save(ctx.deps.storage, &ctx.info.sender, &user)?;
+
+        Ok(())
+    }
+
+    /// Updates the local stake for unstaking from any contract
+    fn unstake(&self, ctx: &mut ExecCtx, owner: String, amount: Coin) -> Result<(), ContractError> {
```
ethanfrey, line 300: Inconsistency to take Uint128 in stake and Coin in unstake.

These are both helpers, so fine to do either, and enforce denom either in caller or in this one. Just nice to be consistent

contracts/vault/src/contract.rs
```diff
@@ -123,19 +246,90 @@ impl VaultContract<'_> {
...
+
+        let mut user = self.users.load(ctx.deps.storage, &owner)?;
+
+        // Max lien has to be recalculatd from scratch; the just released lien
+        // is already decreased in cache
+        user.max_lien = self
+            .liens
+            .prefix(&owner)
+            .range(ctx.deps.storage, None, None, Order::Ascending)
+            .try_fold(Uint128::zero(), |max_lien, lien| {
```
ethanfrey, line 324: style:

`...range().map(|(, lein)| lein.amount).max()` would be cleaner. Or does `Uint128` not implement `Ord`?

ethanfrey: Oh cuz you get `Result<(_, Lein), _>`, right? 

Guess there isn't a better way

contracts/vault/src/contract.rs
```diff
@@ -63,56 +77,165 @@ impl VaultContract<'_> {
...
 
     /// This sends actual tokens to the local staking contract
     #[msg(exec)]
     fn stake_local(
         &self,
-        _ctx: ExecCtx,
+        mut ctx: ExecCtx,
         // amount to stake on that contract
-        amount: Coin,
+        amount: Uint128,
```
ethanfrey, line 169: Same comment with Coin vs Uint128 as above

contracts/vault/src/contract.rs
```diff
@@ -63,56 +77,165 @@ impl VaultContract<'_> {
...
+        let account = Addr::unchecked(account);
+
+        let claims = self
+            .liens
+            .prefix(&account)
+            .range(ctx.deps.storage, None, None, Order::Ascending)
+            .map(|lien| {
+                lien.map(|(lienholder, lien)| LienInfo {
+                    lienholder: lienholder.into(),
+                    amount: lien.amount,
```
ethanfrey, line 211: Note: do you think we should change API to include other info from lein? (eg max_slashable)

contracts/vault/src/contract.rs
```diff
@@ -149,33 +343,53 @@ pub fn reply(deps: DepsMut, _env: Env, reply: Reply) -> Result<Response, Contrac
...
         // address of the user who originally called stake_remote
         owner: String,
         // amount to unstake on that contract
         amount: Coin,
     ) -> Result<Response, ContractError> {
-        let _ = (owner, amount);
-        todo!()
+        nonpayable(&ctx.info)?;
+
+        self.unstake(&mut ctx, owner.clone(), amount.clone())?;
```
ethanfrey, line 362: Oh, weird, so stake takes owner in ctx and leinholder as arg.
And unstake takes leinholder in ctx and owner in arg.

I guess this is actually fine, but please make that design very explicit above both methods (I assumed them to be mirror images of each other)

ethanfrey (commented):
In general, looks like solid implementation of the methods.

I added some comments on some smaller items. Biggest issue is changing `Coin` to `Uint128` some places in public API. (Please revert that and we can have a discussion outside of the PR if you feel really strongly)

Looking forward to seeing a nice example of using Sylvia helpers and mocks, which can inspire writing other contracts

ethanfrey (commented):
In general, looks like solid implementation of the methods.

I added some comments on some smaller items. Biggest issue is changing `Coin` to `Uint128` some places in public API. (Please revert that and we can have a discussion outside of the PR if you feel really strongly)

Looking forward to seeing a nice example of using Sylvia helpers and mocks, which can inspire writing other contracts

contracts/vault/src/contract.rs
hashedone: You are right. I actually wanted to ask about that, and it slipped out of my head. It is basically a trivial change, I will apply.

contracts/vault/src/contract.rs
hashedone: In Sylvia, the `Remote` type (the remote contract basically) implements `AsRef<Addr>`, which means you just call here `contract.as_ref()`. Really didn't want to mess with that in the temporary helper struct knowing there are two more PRs in Sylvia to finish that, so I suppose the beginning of next week is the latest when we have queriers.

contracts/vault/src/contract.rs
hashedone: Ok, here I wanted to write a longer comment later after I finish the tests.

Answering only your question, I don't see any more data to add. In particular - `max_slashable` is queryable from the lienholder, which is returned here. Querying from the front end does not matter, as it can spam local nodes with queries with no additional gas cost; I see no reason for redundancy. We could add it as it is cached, so if someone needs those data, it can save some gas for additional queries, but unless there is a use case, I don't see a reason to.

But what makes me anxious is the linear complexity of this query. I don't mid-querying from API, but it seems like a bad practice if some contract would like to query that. I get that it would probably be a small amount of lienholder per user, but I don't see any hard cap. It seems off to me.

I would prefer to completely get rid of returning claims and add another query, `account_claims` for example returning those claims (possibly paginated). I don't mind adding some more data (while I also don't see a reason for that). Thoughts?

contracts/vault/src/contract.rs
hashedone: I was thinking about this, but it gets clunky - this is the only place to modify it, and it is kind of related to how this function work - changing `max_lien` without `total_slashable` or the other way around does not make sense - therefore if there is a method it should be a single one doing those things together. However then, the method body is not evident without the context of what happens here (it doesn't isolate particular logic), and because of that, I actually decided to leave it this way.

contracts/vault/src/contract.rs
hashedone: Kind of agree, but in unstake I wanted to include common logic on verification of denom, and in stake I wanted to avoid loading denom from the state (pointless gas waste). As I agree with consistency argument, I find this particular choice proper.

contracts/vault/src/contract.rs
hashedone: There is one with itertools, but I don't like this dependency - it is heavy and not that useful. I know I am very biased for folds, but I will use `max` if it would just work ;)

contracts/vault/src/contract.rs
hashedone: It looks pretty much obvious to me - I might not understand this properly, but release functions are called by the staker contracts, while stake by funds owner - that is what I understood from docs and code.

I will make arg names bit better.

contracts/vault/src/state.rs
```diff
@@ -0,0 +1,55 @@
...
+use mesh_apis::local_staking_api::LocalStakingApiHelper;
+
+#[cw_serde]
+pub struct Config {
+    /// The denom we accept for staking (only native tokens)
+    pub denom: String,
+}
+
+#[cw_serde]
+pub struct LocalStaking {
```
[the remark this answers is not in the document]
hashedone: The reason is that they are almost always used separately, especially denom is used all over the place, while local staking is only when dealing with... local staking (and not always needs denom).

Then additional reason - local staking cannon be set up in instantiation because we need to wait for a response - so it has to be filled with dummy data. It is not like real problem, but having it "fixed" for free seems like benefit to me.

contracts/vault/src/contract.rs
```diff
@@ -123,19 +255,96 @@ impl VaultContract<'_> {
         let init_data = parse_instantiate_response_data(&reply.data.unwrap())?;
         let local_staking = Addr::unchecked(init_data.contract_address);
 
-        // we want to calculate the slashing rate on this contract and store it locally...
+        // As we control the local staking contract it might be better to just raw-query it
+        // on demand instead of duplicating the data.
```
JakeHartnell, line 259: Is this comment still relevant?

contracts/vault/src/contract.rs
JakeHartnell: My two cents, think consistency is good and from my point of view `Coin` most closely matches the native staking API. Moreover, it seems a bit more future proof? What if there is someday a chain that allows staking of many different assets? Not trying to over-engineer for that, but ideally we want a solid and stable API that can evolve into V2, V3, etc. without breaking changes.

contracts/vault/src/contract.rs
ethanfrey: > I would prefer to completely get rid of returning claims and add another query, account_claims for example returning those claims (possibly paginated). I don't mind adding some more data (while I also don't see a reason for that). Thoughts?

Yes, that is a very good idea, especially considering the contract query use case

contracts/vault/src/contract.rs
ethanfrey: When I think about it, I see why. I just assumed symmetry and the vague argument names didn't help. I understood at a second read, but this could be clearer

contracts/vault/src/contract.rs
ethanfrey: Fair enough.
When I see the other set method that iterates over all to find max, this becomes more clear why it is not a method

contracts/vault/src/contract.rs
ethanfrey: What Jake said is very aligned with what @alpe said earlier.

Let's just keep it Coin for consistency and future proof.
One more denom field to parse and check is not a big deal.

contracts/vault/src/contract.rs
hashedone: Ok, then I'll add it. Additionally, I would add a query over all users which is also missing.

contracts/vault/src/contract.rs
hashedone: I left it here mostly because the local staking contract is not yet implemented and it is not clear if we can use a raw query for that unless it is done - and I wanted to leave some reminders in the code. While it's in development

contracts/native-staking/src/contract.rs
```diff
@@ -125,102 +125,113 @@ pub fn reply(deps: DepsMut, _env: Env, reply: Reply) -> Result<Response, Contrac
...
-                // Send stake message with funds to the proxy contract
-                let msg =
-                    to_binary(&mesh_native_staking_proxy::contract::ExecMsg::Stake { validator })?;
-                let wasm_msg = WasmMsg::Execute {
-                    contract_addr: proxy_addr.into(),
-                    msg,
-                    funds: ctx.info.funds,
-                };
-                Ok(Response::new().add_message(wasm_msg))
+mod local {
```
jawoznia, line 128: Just suggestion.
I would change this module to separate file. It would avoid nested scope and also for me at least it makes sense to divide the code into logical parts. Not sure if everyone working on this project would agree.

jawoznia (commented):
Nice usage of Sylvia

contracts/native-staking/src/contract.rs
hashedone: I hesitate about this only because this PR is not related to native-staking. On the other hand, I had to change it anyway to align to Sylvia 0.4.2 in the process, so splitting it not to forget later seems to be ok.

contracts/vault/src/contract.rs
ethanfrey: Better argument names. Helps

contracts/vault/src/lib.rs
```diff
@@ -1,39 +1,6 @@
 pub mod contract;
 pub mod error;
-mod types;
-
-#[allow(dead_code)]
-#[cfg(not(any(feature = "library", tarpaulin_include)))]
```
ethanfrey, line 6: Nice, all this is autogenerated now? 👍

contracts/vault/src/msg.rs
```diff
@@ -40,25 +14,23 @@ pub struct StakingInitInfo {
...
 
 #[cw_serde]
-pub struct Lien {
+pub struct LienInfo {
     pub lienholder: String,
     pub amount: Uint128,
 }
+
+#[cw_serde]
+pub struct ConfigResp {
```
ethanfrey, line 51: Nit: Why ConfigResp not ConfigResponse? (not really important, but standardization is nice)

ethanfrey (approved):
Nice one. 
I still have to review the test code, but wanted to give feedback on the contract code.
It looks good. A few minor points to update but with those, good to merge.

contracts/native-staking/src/contract.rs
```diff
@@ -125,102 +125,113 @@ pub fn reply(deps: DepsMut, _env: Env, reply: Reply) -> Result<Response, Contrac
     }
 }
 
-#[contract]
```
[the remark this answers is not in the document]
hashedone: I will update Sylvia's README with this info.

The thing is, Sylvia generates a private module with every contract and interface call, and it is necessary to avoid collision of those modules. It doesn't need to be a separate file, its own module is enough, but it is typically clearer anyway. Naming of the module does not matter.

contracts/vault/src/lib.rs
hashedone: Yup

contracts/vault/src/msg.rs
hashedone: Probably slipped. I am used to using `Resp` in many places and didn't think about it too much. Will unify.

contracts/native-staking/src/contract.rs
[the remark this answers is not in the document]
ethanfrey: Yeah,I figured something like that (this is only for `mt`, right?)

If there are some docs to mention to embed each trait implementation in a `mod { }` block, that is fine

contracts/vault/src/lib.rs
ethanfrey: How do you do this, so I can remove it from other contracts?

contracts/vault/src/contract.rs
maurolacy: Nice construct.

contracts/vault/src/contract.rs
maurolacy: Was thinking about the security implications of this, and couldn't find one. Nice design.

maurolacy (commented):
Reviewed all but tests. Looks good, just some minor details.

contracts/vault/src/lib.rs
hashedone: You just add `sylvia::entry_points` attribute (typcally after `not(feature = "library")` guard).

Keep in mind that error specification also changed to its own `error` attribute.

contracts/vault/src/contract.rs
```diff
@@ -63,56 +78,202 @@ impl VaultContract<'_> {
...
-        todo!()
+        nonpayable(&ctx.info)?;
+
+        let config = self.config.load(ctx.deps.storage)?;
+
+        let contract = ctx.deps.api.addr_validate(&contract)?;
+        let contract = CrossStakingApiHelper(contract);
+        let slashable = contract.max_slash(ctx.deps.as_ref())?;
+
+        let denom = self.config.load(ctx.deps.storage)?.denom;
```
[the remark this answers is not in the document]
hashedone: Obviously, idk how did I doubled it

contracts/vault/src/contract.rs
```diff
@@ -63,56 +78,202 @@ impl VaultContract<'_> {
...
         amount: Coin,
         // action to take with that stake
         msg: Binary,
     ) -> Result<Response, ContractError> {
-        let _ = (amount, msg);
-        todo!()
+        nonpayable(&ctx.info)?;
+
+        let config = self.config.load(ctx.deps.storage)?;
+        ensure!(
```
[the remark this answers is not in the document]
hashedone: @maurolacy good point

Issue #10, opened 2023-04-24
Implement local staking module (simple version)


ethanfrey:
Build and test a contract based on [this documentation](https://github.com/osmosis-labs/mesh-security/blob/main/docs/provider/LocalStaking.md)

It should be able to:
* Accept stake from the vault
* Stake it to any validator of your choice
* Withdraw their share of rewards (these can be batch withdrawn from the staking contract and distributed locally via cw means)
* (Fully or Partially) Unstake from validator
* Once unbonding period is over, send tokens and release claim from the Vault

Future addition (not this issue):
* Allow stakers to vote #8 (overriding validator)

ethanfrey:
Half of this is done.
The rest of it (including #8) should be done in #34

PR #31, opened 2023-05-23
Local native staking impl

maurolacy:
WIP for #10.

TODO:

- ~~Config query~~. (done)
- ~~Max slashing query~~. (done)
- ~~Receive stake~~. (done)
- ~~Release proxy stake~~. (done)
- ~~Proxy by owner query~~. (done)
- ~~Owner by proxy query (uses reverse map)~~. (done)
- Tests.

### DIFF
--- a/contracts/native-staking/src/contract.rs
+++ b/contracts/native-staking/src/contract.rs
@@ -156,13 +156,12 @@ impl LocalStakingApi for NativeStakingContract<'_> {
             }
             Some(proxy_addr) => {
                 // Send stake message to the proxy contract
-                let msg = WasmMsg::Execute {
+                let wasm_msg = WasmMsg::Execute {
                     contract_addr: proxy_addr.into(),
-                    msg,
+                    msg, // StakeMsg
                     funds: ctx.info.funds,
                 };
-                let sub_msg = SubMsg::new(msg);
-                Ok(Response::new().add_submessage(sub_msg))
+                Ok(Response::new().add_message(wasm_msg))
             }
         }
     }
--- a/contracts/native-staking/src/contract.rs
+++ b/contracts/native-staking/src/contract.rs
@@ -212,7 +212,8 @@ impl NativeStakingCallback for NativeStakingContract<'_> {
         // Assert funds are passed in
         let _paid = must_pay(&ctx.info, &cfg.denom)?;
 
-        // Look up account owner by proxy address (info.sender)
+        // Look up account owner by proxy address (info.sender). This asserts the caller is a valid
+        // proxy
         let owner_addr = self
             .owner_by_proxy
             .load(ctx.deps.storage, &ctx.info.sender)?;
--- a/contracts/native-staking/src/contract.rs
+++ b/contracts/native-staking/src/contract.rs
@@ -157,7 +157,7 @@ impl LocalStakingApi for NativeStakingContract<'_> {
             .may_load(ctx.deps.storage, &owner_addr)?
         {
             None => {
-                // Instantiate proxy contract and send stake message, with reply handling on success
+                // Instantiate proxy contract and send funds to stake, with reply handling on success
                 let msg = to_binary(&mesh_native_staking_proxy::contract::InstantiateMsg {
                     denom: cfg.denom,
                     owner: owner.clone(),
@@ -171,11 +171,10 @@ impl LocalStakingApi for NativeStakingContract<'_> {
                     label: format!("LSP for {owner}"),
                 };
                 let sub_msg = SubMsg::reply_on_success(wasm_msg, REPLY_ID_INSTANTIATE);
-                let owner_data = to_binary(&OwnerMsg { owner })?;
-                Ok(Response::new().add_submessage(sub_msg).set_data(owner_data))
+                Ok(Response::new().add_submessage(sub_msg))
             }
             Some(proxy_addr) => {
-                // Send stake message to the proxy contract
+                // Send stake message with funds to the proxy contract
                 let msg =
                     to_binary(&mesh_native_staking_proxy::contract::ExecMsg::Stake { validator })?;
                 let wasm_msg = WasmMsg::Execute {
--- a/contracts/native-staking/src/contract.rs
+++ b/contracts/native-staking/src/contract.rs
@@ -135,7 +135,6 @@ impl LocalStakingApi for NativeStakingContract<'_> {
 
         // Parse message to find validator to stake on
         let StakeMsg { validator } = from_slice(&msg)?;
-        // TODO?: Validate validator address
         let _ = validator;
 
         let owner_addr = ctx.deps.api.addr_validate(&owner)?;
@@ -149,7 +148,7 @@ impl LocalStakingApi for NativeStakingContract<'_> {
                     code_id: cfg.proxy_code_id,
                     msg,
                     funds: ctx.info.funds,
-                    label: format!("LSP for {owner}"), // FIXME: Check / cap label length
+                    label: format!("LSP for {owner}"),
                 };
                 let sub_msg = SubMsg::reply_on_success(msg, REPLY_ID_INSTANTIATE);
                 let owner_data = to_binary(&OwnerMsg { owner })?;
--- a/contracts/native-staking/src/contract.rs
+++ b/contracts/native-staking/src/contract.rs
@@ -24,7 +24,6 @@ pub const REPLY_ID_INSTANTIATE: u64 = 2;
 pub const MAX_SLASH_PERCENTAGE: u64 = 10;
 
 pub struct NativeStakingContract<'a> {
-    // TODO
     config: Item<'a, Config>,
     /// Map of proxy contract address by owner address
     proxy_by_owner: Map<'a, &'a Addr, Addr>,
@@ -91,21 +90,27 @@ impl NativeStakingContract<'_> {
     #[msg(query)]
     fn proxy_by_owner(
         &self,
-        _ctx: QueryCtx,
+        ctx: QueryCtx,
         owner: String,
     ) -> Result<ProxyByOwnerResponse, ContractError> {
-        let _ = owner;
-        todo!()
+        let owner_addr = ctx.deps.api.addr_validate(&owner)?;
+        let proxy_addr = self.proxy_by_owner.load(ctx.deps.storage, &owner_addr)?;
+        Ok(ProxyByOwnerResponse {
+            proxy: proxy_addr.to_string(),
+        })
     }
 
     #[msg(query)]
     fn owner_by_proxy(
         &self,
-        _ctx: QueryCtx,
+        ctx: QueryCtx,
         proxy: String,
     ) -> Result<OwnerByProxyResponse, ContractError> {
-        let _ = proxy;
-        todo!()
+        let proxy_addr = ctx.deps.api.addr_validate(&proxy)?;
+        let owner_addr = self.owner_by_proxy.load(ctx.deps.storage, &proxy_addr)?;
+        Ok(OwnerByProxyResponse {
+            owner: owner_addr.to_string(),
+        })
     }
 }
 
--- a/contracts/native-staking/src/contract.rs
+++ b/contracts/native-staking/src/contract.rs
@@ -26,7 +26,10 @@ pub const MAX_SLASH_PERCENTAGE: u64 = 10;
 pub struct NativeStakingContract<'a> {
     // TODO
     config: Item<'a, Config>,
-    proxies: Map<'a, &'a Addr, Addr>,
+    /// Map of proxy contract address by owner address
+    proxy_by_owner: Map<'a, &'a Addr, Addr>,
+    /// Reverse map of owner address by proxy contract address
+    owner_by_proxy: Map<'a, &'a Addr, Addr>,
 }
 
 #[contract]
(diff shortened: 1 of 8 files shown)


contracts/native-staking/src/contract.rs
```diff
@@ -49,8 +60,24 @@ impl NativeStakingContract<'_> {
...
+    fn reply_init_callback(
+        &self,
+        deps: DepsMut,
+        reply: SubMsgResponse,
+    ) -> Result<Response, ContractError> {
+        let init_data = parse_instantiate_response_data(&reply.data.unwrap())?;
+
+        // Associate staking proxy with owner address
+        let proxy_addr = Addr::unchecked(init_data.contract_address);
+        let owner_data: OwnerMsg = from_slice(&init_data.data.unwrap())?; // TODO?: Check for None
```
hashedone, line 76: I would avoid unwrapping. Better to raise some missing data or invalid data errors. On the other hand, it is the response to instantiation, so we control the contract ourselves, so neither of those should fire (especially the first one), still - it is called once, and checking (at least the second one) really won't hurt gas-wise, and will report a bug better.

contracts/native-staking/src/contract.rs
```diff
@@ -49,8 +60,24 @@ impl NativeStakingContract<'_> {
...
+        let owner_addr = Addr::unchecked(owner_data.owner);
```
hashedone, line 77: This I don't necessarily like. If it would be the address of instantiated address returned from instantiation (here: `proxy_addr`) all would be fine, as this one is guaranteed to be valid by the chain. However, here you have an address set on the response by some external contract, and in case of a bug or accidentally messing up the code id of instantiated contract (maybe some similar contract that just ignored unwanted message fields or malicious contract with code provided by someone) and it can be invalid.

I know that it still can be valid, but not a proper address technically, which we cannot protect from. Still, for better bug detection/reporting, I would prefer to see an invalid address error here than when I try to communicate with it later in the contract.

hashedone (commented):
Just some comments after a quick look - I didn't do it carefully, and completely ignored parts that still look to be in progress, but have some suggestions about minorities that might be already useful.

contracts/native-staking/src/contract.rs
maurolacy: You mean, add address validation?

The point here is, that this reply handler can only be invoked by this contract itself. I mean, this contract should initiate the entire process, and therefore, it controls the reply's contents, including the payload; as long as this contract is requesting a reply from a known party, that is.

contracts/native-staking/src/contract.rs
ethanfrey: Same thing. I prefer `data.ok_or(ContractError::NoInstatiateData)?`

contracts/native-staking/src/contract.rs
```diff
@@ -1,22 +1,32 @@
...
-use crate::types::{Config, ConfigResponse, OwnerByProxyResponse, ProxyByOwnerResponse, StakeMsg};
+use crate::msg::{ConfigResponse, OwnerByProxyResponse, OwnerMsg, ProxyByOwnerResponse, StakeMsg};
+use crate::state::Config;
 
 pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+pub const REPLY_ID_INSTANTIATE: u64 = 2;
+
+// TODO: Hardcoded for now. Revisit for v1.
```
ethanfrey, line 23: This is fair enough.

contracts/native-staking/src/contract.rs
```diff
@@ -49,8 +60,24 @@ impl NativeStakingContract<'_> {
     }
 
     #[msg(query)]
-    fn config(&self, _ctx: QueryCtx) -> Result<ConfigResponse, ContractError> {
-        todo!()
+    fn config(&self, ctx: QueryCtx) -> Result<ConfigResponse, ContractError> {
+        self.config.load(ctx.deps.storage).map_err(Into::into)
```
ethanfrey, line 64: Nit: I prefer `Ok(xyz()?)` expression to `xyz().map_err(Into::into)`
They have the same effect, neither is "better". Just my style for what it is worth.

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +125,55 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
         ensure_eq!(cfg.vault, ctx.info.sender, ContractError::Unauthorized {});
 
-        // assert funds passed in
+        // Assert funds are passed in
         let _paid = must_pay(&ctx.info, &cfg.denom)?;
 
-        // parse message to find validator to stake on
+        // Parse message to find validator to stake on
         let StakeMsg { validator } = from_slice(&msg)?;
+        // TODO?: Validate validator address
```
ethanfrey, line 137: No way to validate (different format, so deps.api can't verify).
Just pass along and it will be validated when staking.

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +125,55 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
+                let sub_msg = SubMsg::reply_on_success(msg, REPLY_ID_INSTANTIATE);
+                let owner_data = to_binary(&OwnerMsg { owner })?;
+                Ok(Response::new().add_submessage(sub_msg).set_data(owner_data))
+            }
+            Some(proxy_addr) => {
+                // Send stake message to the proxy contract
+                let msg = WasmMsg::Execute {
+                    contract_addr: proxy_addr.into(),
+                    msg,
+                    funds: ctx.info.funds,
```
ethanfrey, line 162: Ah, clever, you never need to look at `_paid` value above, as you just pass along funds

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +125,55 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
+        // Look up if there is a proxy to match. Instantiate or call stake on existing
+        match self.proxies.may_load(ctx.deps.storage, &owner_addr)? {
+            None => {
+                // Instantiate proxy contract and send stake message, with reply handling on success
+                let msg = WasmMsg::Instantiate {
+                    admin: Some(ctx.env.contract.address.into()),
+                    code_id: cfg.proxy_code_id,
+                    msg,
+                    funds: ctx.info.funds,
+                    label: format!("LSP for {owner}"), // FIXME: Check / cap label length
```
ethanfrey, line 151: This is fine. Labels are like 128 limit. One address is circa 60.

2 addresses in a label may be problematic (if they reference contracts not external accounts)

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +125,55 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
+            }
+            Some(proxy_addr) => {
+                // Send stake message to the proxy contract
+                let msg = WasmMsg::Execute {
+                    contract_addr: proxy_addr.into(),
+                    msg,
+                    funds: ctx.info.funds,
+                };
+                let sub_msg = SubMsg::new(msg);
+                Ok(Response::new().add_submessage(sub_msg))
```
ethanfrey, line 165: Nit: `Ok(Response::new().add_message(msg))`
does the exact same thing without the line above

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +125,55 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
+        let owner_addr = ctx.deps.api.addr_validate(&owner)?;
+
+        // Look up if there is a proxy to match. Instantiate or call stake on existing
+        match self.proxies.may_load(ctx.deps.storage, &owner_addr)? {
+            None => {
+                // Instantiate proxy contract and send stake message, with reply handling on success
+                let msg = WasmMsg::Instantiate {
+                    admin: Some(ctx.env.contract.address.into()),
+                    code_id: cfg.proxy_code_id,
+                    msg,
```
ethanfrey, line 149: You don't want to blindly pass msg.
You need to build a new message (with denom, owner, validator). 
See [proxy design](https://github.com/osmosis-labs/mesh-security/blob/main/contracts/native-staking-proxy/src/contract.rs#L30-L38)

2 points:
1. It would be nice to have auto-generated helpers to call this type-safe. I mean `native_staking_proxy::contract::InstantiateMsg` _should_ be the right type. I was proposing to make helper functions to build this next level. (And @hashedone can give a tip on how to call type-safe)
2. If you think `denom` is unneeded, let's remove it from the message. We can probably get it by querying `StakingQuery::BondDenom{}` or such. But if removed, I'd remove it from the instantiate method of this contract as well.

ethanfrey (commented):
Nice start, added some comments.

contracts/native-staking/src/contract.rs
JakeHartnell: Yeah, went through a [long discussion with this previously on the prototype](https://github.com/CosmWasm/mesh-security-hackathon/issues/42). FWIW, some newer Cosmos chains like Penumbra are also ditching `valoper` addresses (not to mention other non-SDK cosmwasm VMs like Composable), so not sure it's worth validating in the contract as it's done on the SDK or native chain code.

contracts/native-staking/src/contract.rs
maurolacy: Good catch. Fixed it.

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +139,62 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
+                let wasm_msg = WasmMsg::Instantiate {
+                    admin: Some(ctx.env.contract.address.into()),
+                    code_id: cfg.proxy_code_id,
+                    msg,
+                    funds: ctx.info.funds,
+                    label: format!("LSP for {owner}"),
+                };
+                let sub_msg = SubMsg::reply_on_success(wasm_msg, REPLY_ID_INSTANTIATE);
+                let owner_data = to_binary(&OwnerMsg { owner })?;
+                Ok(Response::new().add_submessage(sub_msg).set_data(owner_data))
```
ethanfrey, line 175: Why setting data here? I mean the caller knows `owner` already.

If you want to return the `proxy` address, that might make sense to save a later query. But totally optional as it can be queried.

contracts/native-staking/src/contract.rs
```diff
@@ -88,30 +139,62 @@ impl LocalStakingApi for NativeStakingContract<'_> {
...
-        let _ = validator;
-
-        let _ = owner;
 
-        // look up if there is a proxy to match
-        // instantiate or call stake on existing
-        todo!();
+        let owner_addr = ctx.deps.api.addr_validate(&owner)?;
+
+        // Look up if there is a proxy to match. Instantiate or call stake on existing
```
ethanfrey, line 154: Method looks good in general

contracts/native-staking/src/contract.rs
```diff
@@ -123,10 +206,21 @@ impl NativeStakingCallback for NativeStakingContract<'_> {
...
-        // look up proxy address (info.sender) to account owner
-        // send these tokens to vault contract, using release_local_stake method
-        todo!()
+    fn release_proxy_stake(&self, ctx: ExecCtx) -> Result<Response, Self::Error> {
+        let cfg = self.config.load(ctx.deps.storage)?;
+
+        // Assert funds are passed in
+        let _paid = must_pay(&ctx.info, &cfg.denom)?;
+
+        // Look up account owner by proxy address (info.sender)
```
ethanfrey, line 215: Add a note that this asserts the caller is a valid proxy

ethanfrey (approved):
First of all, nice work.

It is unfortunate, that this is based on vault. It would be good to get smaller independent PRs merged often to main.

In particularly now, we need:
1. Tests on this and #30 are both blocked on a new Sylvia release (likely tomorrow)
2. I am unclear when vault will be merged
3. native-staking-proxy needs a new PR

I would actually be happy to have both merged when they compile and pass CI, but missing tests. Then do tests in a new PR (I say this here as there is sylvia changes needed for tests and getting them in now may avoid a lot of conflicts and contention)

In short:
* Please rebase/cherry-pick so this is directly off of main
* Let's merge it
* Start a new PR adding native-staking-proxy (without tests)
* When vault lands (with proof of working Sylvia multi-test reply helpers), make new PRs adding tests

contracts/native-staking/src/contract.rs
maurolacy: The reply handler needs it (so as to properly populate the maps). But good point, we don't need to set here, but in the staking proxy instantiate return message. And it can get it from the `owner` field, above. Simpler is better.

maurolacy:
> It is unfortunate, that this is based on vault. It would be good to get smaller independent PRs merged often to main.

I did just to take a look / copy code from it.
 
> 
> * Please rebase/cherry-pick so this is directly off of main

Just created a fresh branch from `main`, and cherry picked the relevant commit range.

> * Let's merge it
> * Start a new PR adding native-staking-proxy (without tests)
> * When vault lands (with proof of working Sylvia multi-test reply helpers), make new PRs adding tests

👍🏼

maurolacy:
Was taking a look at the CI errors, and can't make sense of them. Do we merge anyway? Will take a closer look tomorrow, in any case.

maurolacy:
> Was taking a look at the CI errors, and can't make sense of them. Do we merge anyway? Will take a closer look tomorrow, in any case.

Solved it by `cargo update`ing sylvia.

ethanfrey:
Great, thanks for the cleanup.
I will rebase on top.

PR #33, opened 2023-05-24
Define Consumer Contracts

ethanfrey:
This will start some contract code from the consumer side. The intention is to clarify some interfaces in the documentation by means of sample code. But also, to work towards a minimal working virtual staking *contract* that can be used for system-level integrations with the virtual staking *SDK module*.

- [x] Define new interfaces between these contracts
- [x] Create minimal gov-set price-feed contract
- [x] Start native staking contract (no reward withdrawal)
- [x] Provide stub for the converter contract


contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    #[msg(query)]
+    fn config(&self, ctx: QueryCtx) -> Result<ConfigResponse, ContractError> {
+        Ok(self.config.load(ctx.deps.storage)?.into())
+    }
+
+    fn handle_epoch(
+        &self,
+        deps: DepsMut<VirtualStakeCustomQuery>,
```
hashedone, line 62: Ouch, I am not sure if it is properly supported in Sylvia to use custom queries, it was postponed. I will double-check, and if not, we will push for delivering it asap.

@jawoznia

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -0,0 +1,159 @@
...
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub struct VirtualStakingContract<'a> {
+    config: Item<'a, Config>,
+    // Amount of tokens that have been requested to bond to a validator
+    // (Sum of bond minus unbond requests). This will only update actual bond on epoch changes.
+    bond_requests: Map<'a, &'a str, Uint128>,
+    // This is how much was bonded last time (validator, amount) pairs
+    bonded: Item<'a, Vec<(String, Uint128)>>,
```
hashedone, line 23: It seems very off. I assume `validator` is unique, and if so - why is it not `Map<&Addr, Uint128>`?

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -0,0 +1,159 @@
...
+use crate::types::{Config, ConfigResponse};
+
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub struct VirtualStakingContract<'a> {
+    config: Item<'a, Config>,
+    // Amount of tokens that have been requested to bond to a validator
+    // (Sum of bond minus unbond requests). This will only update actual bond on epoch changes.
+    bond_requests: Map<'a, &'a str, Uint128>,
```
hashedone, line 21: This key should probably be an `&Addr`

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -0,0 +1,159 @@
...
+        let total_request: Uint128 = requests.iter().map(|(_, v)| v).sum();
+        if total_request > bond.cap.amount {
+            // TODO: normalize the list of requests to match the cap
+            todo!();
+        }
+
+        // TODO: Look at existing list and calculate differences for bond/unbond calls
+
+        // Save the new values
+        self.bonded.save(deps.storage, &requests)?;
```
hashedone, line 87: Collecting all the requests, then summing them, then ignoring those out of range, and store after transformation seems waaaay off to me. I assume that normalizing requests is done to prevent wasting gas on transforming all the requests, but probably more gas is wasted by collecting all the requests, and always fetching/storing all the bonded items.

The algorithm should probably go ass follows:

The requests should never be collected, instead they should be transformed fold-like keeping the transformed items, and total request (both as `Iterator::scan` items). The reason for using `scan` instead of fold is, that `fold` is eager, and `scan` is lazy - after scan we can use `take_while`, to stop transforming after the cap is reached.

I would go even deeper, and in the `scan` I would just return an iterator over `(request, total)`, then `take_wile` to break when `total` reaches `cap`, then `map` to transform items and update them immediately in bonded (which should be `Map` for that.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -0,0 +1,159 @@
...
+        Ok(Response::new())
+    }
+}
+
+#[contract]
```
hashedone, line 92: #[messages(...)]` missing

contracts/consumer/virtual-staking/src/types.rs
hashedone, line 1: I would split it into the `state.rs` and `msg.rs` - note that you split it with comments, which means you see two separated logical entities, why not split them into modules? The module cannot be too small if it is logically an entity.

hashedone (commented):
Some comments for current state.

contracts/consumer/virtual-staking/src/contract.rs
ethanfrey: Good thing is sudo is outside of Sylvia so I can get this to compile and run for system level tests (uploading to a chain), even if I can't for multitest yet

contracts/consumer/virtual-staking/src/contract.rs
ethanfrey: The key is a validator address (specifically `valoper`), which due to some weird architectural decision of the SDK uses a different bech32 prefix than the normal accounts, although it actually works the same way 🤷‍♂️ 

We have to treat as String without validating.

contracts/consumer/virtual-staking/src/contract.rs
ethanfrey: Happy for better design, but let me finish this, then improve. I need to:

1. Get all bond requests
2. Sum the total amount
3. If the sum <= max_cap then use collected requests as is
4. If the sum > max_cap,
  a. calculate multiplier Decimal(max_cap / sum)
  b. multiply every element of the collected requests in place.
5. Find diff between collected (normalized) requests and last bonding amounts (which go up, which down).
6. Transform diff into unbond and bond requests, sorting so all unbond happen first

The largest optimization I can see is collecting (in 1) into a {Hash,BTree}Map as we will likely need that later. 

We don't cut off at the cap, we multiply everyone down. We need to get it all.

The storage load is the most expensive. Highest V is 100-150, so even O(N^2) in memory, which I would like to avoid, may be cheaper than the storage reads

contracts/consumer/virtual-staking/src/contract.rs
ethanfrey: Bonded could be a Map, but the only time we use it is to read / write the entire list in bulk, never accessing one element. Reading 100 elements in an Item is much cheaper than ranging over a Map with 100 entries.

I keep bond_requests as a Map, as most interactions are bond/unbond requests (from IBC) which touch one. And then we only range over it once per epoch.

Since bonded is only read as a whole, this is more efficient.

contracts/consumer/virtual-staking/src/contract.rs
ethanfrey: NB: I rebased on your vault branch to check new Sylvia and it errors on returning `Response<CustomMsg>` from instantiate. (something about `anyhow::Error` so I assume `mt` related. One more thing to check then

ethanfrey:
The basic code is laid out. There are no tests.

Simple Price Feed is simple enough to be tested, and a good place to learn how, but not really needed.
The Virtual Staking contract, which really could use serious tests, is not testable with current Sylvia framework, as most logic happens in sudo entry point with CustomMsg and CustomQuery... [all of which need new Sylvia features to test](https://github.com/osmosis-labs/mesh-security/pull/33#discussion_r1205671859)

However, this is a first step to getting something that @alpe could use in system tests (as the generated wasm should work)

contracts/consumer/virtual-staking/src/contract.rs
hashedone: Yea, both custom queries and custom messages are not supported. We will prioritize those.

Sudo is easy thing to do, for now we can do it in reply-wise and deprecate it when we properly implement them.

ethanfrey:
@hashedone I rebased on vault merge, can you review this (ideally run `cargo build` and `cargo clippy` locally).

The only issue I have is a build warning. Code:

```rust
use mesh_apis::local_staking_api::{self, LocalStakingApi, MaxSlashResponse};

#[contract]
#[messages(local_staking_api as LocalStakingApi)]
impl LocalStakingApi for NativeStakingContract<'_> {
 // ...
}
```

Error:

```
warning: unused import: `self`
 --> contracts/native-staking/src/local_staking_api.rs:6:36
  |
6 | use mesh_apis::local_staking_api::{self, LocalStakingApi, MaxSlashResponse};
  |                                    ^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

I guess the `#messages[...]` block is unused there? And I should remove it?

contracts/consumer/simple-price-feed/src/contract.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::state::Config;
+
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub struct SimplePriceFeedContract<'a> {
+    pub config: Item<'a, Config>,
+}
+
+#[contract]
```
hashedone, line 21: Here also - `entry_points`

contracts/consumer/simple-price-feed/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    #[entry_point]
+    pub fn query(deps: Deps, env: Env, msg: ContractQueryMsg) -> Result<Binary, ContractError> {
+        msg.dispatch(&CONTRACT, (deps, env))
+    }
+}
+
+#[cfg(not(any(feature = "library", tarpaulin_include)))]
+pub use crate::entry_points::*;
```
hashedone, line 44: And then this is also not needed

contracts/consumer/virtual-staking/src/contract.rs
hashedone: Not 100% sure how to approach it. This is too big PR to leave it not merged, but queries/responses would be in Sylvia this/next week (Jan is working on that right now). Sudo messages I can do in one day (not the way replies go, I have another approach for that which will stay relevant). On the other way, I assume it won't work properly so it is merging invalid code probably.

My take: change the reply to "default" for now, and where you query it just create a dummy default response, leaving todo: "Properly query when https://github.com/CosmWasm/sylvia/issues/16 is merged". Then create issue on GH to fix it (to not forget).

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -0,0 +1,250 @@
...
+        // calculate what the delegations should be when we are done
+        let mut requests: Vec<(String, Uint128)> = self
+            .bond_requests
+            .range(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+            .collect::<Result<_, _>>()?;
+        let total_requested: Uint128 = requests.iter().map(|(_, v)| v).sum();
+        if total_requested > max_cap {
+            requests.iter_mut().for_each(|(_, v)| {
+                *v = (*v * max_cap) / total_requested;
+            });
```
hashedone, line 118: I tend to avoid any `for_each` as this is just a for loop; I find it far more readable this way:

```rust
for (_, v) in &mut requests {
  *v = (*v * max_cap) / total_requested;
}
```

contracts/consumer/virtual-staking/src/contract.rs
hashedone: Ahh, ignore it - it is used only in sudo. So you just use sudo this way which is working for now, except not testable at this point. But we will deliver it soon and then change. Good (for now).

hashedone (approved):
Couple comments.

First of all - entry points can be generated with a single line from Sylvia - who commented on it twice, but there are more contracts and that apply to all of them.

The warning about unused modules is surprising, but it is not because `#[messages(...)]` is not used - it is always used, but when `mt` utils are not generated (so no `mt` flag enabled on Sylvia), the macro does not use the provided path. Don't remove the `#[messages(...)]` attribute; it will cause more problems in the future. For now, the best I can think of is adding `#[allow(unused_imports)]` just before the import block which causes this one. We will fix this warning in Sylvia by artificially using the module even when not generating things using it.

I am leaving the approval as I don't see any serious problem so you can merge when you fix or ignore comments.

contracts/consumer/converter/src/contract.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub struct ConverterContract<'a> {
+    pub config: Item<'a, Config>,
+    pub virtual_stake: Item<'a, Addr>,
+}
+
+#[contract]
```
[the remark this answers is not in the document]
ethanfrey: It must be before `#[contract]`. I tried after, got a compiler error about ContractError not implementing `Into<StdError>`.

I just put it before, fine as long as you document it

ethanfrey:
Made requested fixes and rebased on main.
Will merge when this passes CI

PR #35, opened 2023-05-26
Local staking proxy

maurolacy:
WIP for #31.

TODO:

- ~~Stake~~. (done)
- ~~Restake~~. (done)
- ~~Config query~~. (done)
- ~~Vote~~. (done)
- ~~Vote weighted~~. (done)
- ~~Withdraw rewards~~. (done)
- ~~Release unbounded~~. (done)
- Unbonding. (needs API / query support)
- Tests.


maurolacy:
Couple questions here:

1. Saw that I can get the list of all delegations using the standard `StakingQuery::AllDelegations`. Do I use it, or we prefer to keep track of it in the contract, as already implemented? (I guess the former. Will change the code).
2. AFAIK Ttere's no way to get the list of pending unbondings. One way would be to keep track of the `unstake`s, but then I would need to know the unbonding period, to implement something like claims expiration / processing. Can't find a way to query for it.

hashedone (commented):
Overall good, I don't see any logic problem, assuming all is fine.

Would appreciate some tests on that.

maurolacy:
> 
> Would appreciate some tests on that.

Will do / WIP.

maurolacy:
Merging this, to continue working on integrated tests in other branches. Will write tests in an upcoming PR.

Issue #13, opened 2023-04-24
Implement Remote Staking contract (no rewards)


ethanfrey:
* It takes leins from the Vault and sends IBC messages
* User can specify which validator
* User can initiate (partial or full) unbonding - triggers IBC message and local claim window
* When local claim completed, can release (full or partial) lein on the vault
~~* Currently a user must have all tokens staked on one remote validator, and cannot split (Fix for V1 - #12)~~

There is a follow up issue to handle rewards: https://github.com/osmosis-labs/mesh-security/issues/43

ethanfrey:
For first demo version, let's do this:

* Bond message from the vault (leave a TODO where it should send IBC message, but don't implement that)
* Unstake message from owner
  * Move from staking to "unbonding" window. (record expiration)
  * Leave TODO for sending IBC message
* Claim/Withdraw/??? message (pick a better name) from owner
  * Once unstake record is "mature" (expiration has passed), then we allow them to release lien on vault contract

ethanfrey:
The next step can make a "distribute rewards" message callable by anyone. It takes tokens (the remote IBC token, can be specified in init, different than native staking)

This should be a cw2222 style implementation for lazy withdrawl. We need some changes, as we track rewards for each validator separately, thus have a dynamic set of  buckets for withdraw.

We need to specify (validator, user) pair to withdraw rewards. Maybe we can query the validator given the user, so we only expose user (`info.sender`) in the `WithdrawRewards` message. And look up the validator from contract-local storage before calling into the cw2222 code.

hashedone:
To be done:
- [ ] Implement basic bare-bone contract for external staking
- [ ] Add the possibility to choose the validator when staking externally for the first time
- [ ] Test for external staking/unstaking
- [ ] Rewards distribution (cw2222)
- [ ] Tests for rewards distribution

PR #37, opened 2023-05-31
Remote staking contract bare-bones

hashedone:
Addresses #13 

This is just the most basic possible remote staking, with nothing fancy. It meant to address only the first item of https://github.com/osmosis-labs/mesh-security/issues/13#issuecomment-1569538835.

No tests and no validator choice by design - splitting into separate PRs, smaller = easier to review.

Next step: add the possibility of choosing the validator to stake on passing a message when staking (#39).

### DIFF
--- a/contracts/external-staking/src/contract.rs
+++ b/contracts/external-staking/src/contract.rs
@@ -1,7 +1,6 @@
-use cosmwasm_std::{ensure, ensure_eq, Addr, Coin, Order, Response};
+use cosmwasm_std::{coin, ensure, ensure_eq, Addr, Coin, Order, Response};
 use cw2::set_contract_version;
 use cw_storage_plus::{Bounder, Item, Map};
-use cw_utils::Duration;
 use mesh_apis::cross_staking_api;
 use mesh_apis::vault_api::VaultApiHelper;
 use sylvia::contract;
@@ -9,7 +8,7 @@ use sylvia::types::{ExecCtx, InstantiateCtx, QueryCtx};
 
 use crate::error::ContractError;
 use crate::msg::{ConfigResponse, UserInfo, UsersResponse};
-use crate::state::{Config, PendingUnbound, User};
+use crate::state::{Config, PendingUnbond, User};
 
 pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -45,7 +44,7 @@ impl ExternalStakingContract<'_> {
         ctx: InstantiateCtx,
         denom: String,
         vault: String,
-        unbounding_period: Duration,
+        unbonding_period: u64,
     ) -> Result<Response, ContractError> {
         let vault = ctx.deps.api.addr_validate(&vault)?;
         let vault = VaultApiHelper(vault);
@@ -53,7 +52,7 @@ impl ExternalStakingContract<'_> {
         let config = Config {
             denom,
             vault,
-            unbounding_period,
+            unbonding_period,
         };
 
         self.config.save(ctx.deps.storage, &config)?;
@@ -63,8 +62,8 @@ impl ExternalStakingContract<'_> {
         Ok(Response::new())
     }
 
-    /// Schedules tokens for release, adding them to the pending unbounds. After `unbounding_period`
-    /// passes, funds are ready to be released with `claim` call by the user
+    /// Schedules tokens for release, adding them to the pending unbonds. After `unbonding_period`
+    /// passes, funds are ready to be released with `withdraw_unbonded` call by the user
     #[msg(exec)]
     pub fn unstake(&self, ctx: ExecCtx, amount: Coin) -> Result<Response, ContractError> {
         let config = self.config.load(ctx.deps.storage)?;
@@ -87,12 +86,12 @@ impl ExternalStakingContract<'_> {
 
         user.stake -= amount.amount;
 
-        let release_at = config.unbounding_period.after(&ctx.env.block);
-        let unbound = PendingUnbound {
+        let release_at = ctx.env.block.time.plus_seconds(config.unbonding_period);
+        let unbond = PendingUnbond {
             amount: amount.amount,
             release_at,
         };
-        user.pending_unbounds.push(unbound);
+        user.pending_unbonds.push(unbond);
 
         self.users.save(ctx.deps.storage, &ctx.info.sender, &user)?;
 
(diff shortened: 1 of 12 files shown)


contracts/external-staking/src/contract.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub config: Item<'a, Config>,
+    pub users: Map<'a, &'a Addr, User>,
+}
+
+#[cfg_attr(not(feature = "library"), sylvia::entry_points)]
+#[contract]
+#[error(ContractError)]
+#[messages(cross_staking_api as CrossStakingApi)]
+impl ExternalStakingContract<'_> {
+    pub const fn new() -> Self {
```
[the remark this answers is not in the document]
hashedone: No, it can be const, and it works perfectly well being const. It should be const in all cases it can be, as even if you call on-demand in the entry point compiler can optimize it out to the static section of binary if the function is constant without inlining it.

Every function which can be const should be const.

maurolacy (approved):
LGTM. Just some syntax / naming stuff.

contracts/external-staking/src/state.rs
```diff
@@ -0,0 +1,66 @@
...
+        };
+
+        // First item is still not ready for release
+        if !self.pending_unbounds[0].release_at.is_expired(&info) {
+            return;
+        }
+
+        let non_expired_idx = self
+            .pending_unbounds
+            .partition_point(|pending| pending.release_at.is_expired(&info));
```
maurolacy, line 58: Perhaps these two can be combined? Just curious.

contracts/external-staking/src/contract.rs
```diff
@@ -0,0 +1,217 @@
...
+        Ok(resp)
+    }
+
+    /// Queries for user-related info
+    ///
+    /// If user not existin in the system is queried, the default "nothing staken" user
+    /// is returned - also if user is not a valid address.
+    #[msg(query)]
+    pub fn user(&self, ctx: QueryCtx, user: String) -> Result<User, ContractError> {
+        let user = Addr::unchecked(user);
```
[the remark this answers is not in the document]
ethanfrey: Yes, please never use `Addr::unchecked` outside of test code (unless you have a very good reason for some special case, like proto parsing the instantiate reply data that comes from x/wasm)

contracts/external-staking/src/state.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/// All user/account related information
+#[cw_serde]
+#[derive(Default)]
+pub struct User {
+    /// How much tokens user staken and not in unbounding period
+    /// via this contract
+    pub stake: Uint128,
+    /// List of token batches scheduled for unbouding
```
ethanfrey, line 24: Could you add comment this should be sorted from earliest to latest to expire (and then ensure that when adding, and take advantage of the fact when removing)

contracts/external-staking/src/state.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub released: Uint128,
+}
+
+/// Description of tokens in unbounding period
+#[cw_serde]
+pub struct PendingUnbound {
+    /// Tokens scheduled for unbounding
+    pub amount: Uint128,
+    /// Time when tokens are released
+    pub release_at: Expiration,
```
ethanfrey, line 36: `Expiration` is an enum over `Time`, `Height`, `Never`
If we always use `Time`, better to just use `Timestamp` (unix nanoseconds) or `u64` (unix seconds) here.
This makes comparison and other reading logic easier

contracts/external-staking/src/state.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub fn release_pending(&mut self, info: BlockInfo) {
+        // The fact that `pending unbounds are always added to the end, so they are always ordered
+        // is assumed here.
+
+        // Nothing waits for unbound
+        if self.pending_unbounds.is_empty() {
+            return;
+        };
+
+        // First item is still not ready for release
```
ethanfrey, line 51: Nice optimizing checks here

contracts/external-staking/src/contract.rs
```diff
@@ -0,0 +1,217 @@
...
+        let release_at = config.unbounding_period.after(&ctx.env.block);
+        let unbound = PendingUnbound {
+            amount: amount.amount,
+            release_at,
+        };
+        user.pending_unbounds.push(unbound);
+
+        self.users.save(ctx.deps.storage, &ctx.info.sender, &user)?;
+
+        // TODO:
```
ethanfrey, line 99: Yes, this is a `TODO: send IBC message` (just like when staking - those are the two key events we inform the consumer chain of)

Unstaking happens on provider. This is specified in the docs

contracts/external-staking/src/contract.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// By default claims tokens to the message sender, but owner can be specified to claim
+    /// tokens on behalf of another user.
+    ///
+    /// Tokens to be claimed has to be unboud before by calling the `unbound` message and
+    /// waiting the `unbound_period`
+    #[msg(exec)]
+    pub fn claim(
+        &self,
+        ctx: ExecCtx,
+        owner: Option<String>,
```
ethanfrey, line 122: This API confused me a lot.

1. When would we claim on behalf of someone else? Not wrong, but those tokens would go back to the vault from the account that owned them.
2. Why specify amount? We should just do `user. release_pending()` and then send all released virtual tokens to the vault.

Note this is about returning a virtual stake / lein to the vault, not claiming rewards. Maybe it needs a better name (like `release_unbonded`?)

contracts/external-staking/src/contract.rs
```diff
@@ -0,0 +1,217 @@
...
+            config.denom,
+            ContractError::InvalidDenom(config.denom)
+        );
+
+        let mut user = self
+            .users
+            .may_load(ctx.deps.storage, &owner)?
+            .unwrap_or_default();
+
+        ensure!(
```
ethanfrey, line 143: I would change the logic on the following lines:

1. Call `user.release_pending()`
2. Use the amount released and `cfg.denom` for `release_cross_stake`
3. Remove the `released` field on user (which is only an intermediate state) and just return `Uint128` from `release_pending`, using that her to release_cross_stake

contracts/external-staking/src/contract.rs
```diff
@@ -0,0 +1,217 @@
...
+        let user = Addr::unchecked(user);
+        let user = self
+            .users
+            .may_load(ctx.deps.storage, &user)?
+            .unwrap_or_default();
+        Ok(user)
+    }
+
+    #[msg(query)]
+    pub fn users(
```
ethanfrey, line 188: Not sure this is needed, but not bad to have.
It would be good to get feedback from those building the UI on any more queries they want

ethanfrey (commented):
In general the logic looks good, except for claims.
A number of minor nit-picks, it would be good to clean this up, so we can merge it.
I will look at your other PR, but nice to rebase it on a cleaned/merged version of this

contracts/external-staking/src/cross_staking_api.rs
```diff
@@ -0,0 +1,65 @@
+use cosmwasm_std::{ensure_eq, Binary, Coin, Decimal, Response};
```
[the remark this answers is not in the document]
hashedone: I can do it in this PR. I was really considering both options, and I don't mind moving it there. Small enough so contract doesn't overgrow.

contracts/external-staking/src/state.rs
hashedone: And now there is as struggle - I agree with you, but actually, when working with `utils::Duration`, using `Expiration` is easier API-wise. If it would never every be anything but time, then I am ok changing it to `Timestamp`, but also not use `utils::Duration`, but `u64` seconds there to not filter out invalid variants.

Based on this comment I assume it is always time so I'll do the change.

contracts/external-staking/src/contract.rs
hashedone: I don't know. Maybe the owner is completely redundant, I'll remove it. Didn't have a particular UC in my head, just added it. As well as coins. Ill rename it to `withdraw_unbounded`.

contracts/external-staking/src/contract.rs
hashedone: Yes, this is what should happen - it is a bug. I realized it and fixed it (I think) in the following review. I will fix it here to not commit invalid code.

contracts/external-staking/src/contract.rs
hashedone: Very easy to make natural query - not see a reason not to provide it just in case.

contracts/external-staking/src/cross_staking_api.rs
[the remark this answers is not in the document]
hashedone: Ok, change of plans - let me do it in the next PR. Reason: I did the other one and I don't want to mess up the rebase with these big chunks changes.

contracts/external-staking/src/state.rs
hashedone: I just redo it, completely forgot about the `drain` method of vec

PR #38, opened 2023-05-31
Local staking tests

maurolacy:
Native-staking multi-tests.

TODO:

- ~~`instantiate` test~~. (done)
- ~~`max_slash` test~~. (done)
- ~~`receive_stake` test~~. (done)
- ~~`release_proxy_stake` test~~. (done)
- ~~Connect with real impls / remove other contracts stubs when possible~~ (done for vault contract in `release_proxy_stake` test).


hashedone (approved):
All seems good, minor comments.

contracts/native-staking/src/contract.rs
```diff
@@ -65,6 +66,14 @@ impl NativeStakingContract<'_> {
         self.config.load(ctx.deps.storage).map_err(Into::into)
     }
 
+    #[msg(reply)]
+    fn reply(&self, ctx: ReplyCtx, reply: Reply) -> Result<Response, ContractError> {
+        match reply.id {
+            REPLY_ID_INSTANTIATE => self.reply_init_callback(ctx.deps, reply.result.unwrap()),
```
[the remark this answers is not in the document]
maurolacy: I'm afraid I've copied that code verbatim from your `vault` impl 😺. Let's change it in both places, in any case.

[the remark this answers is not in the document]
ethanfrey: Happy for a different idiom there

packages/apis/src/local_staking_api.rs
```diff
@@ -45,10 +45,11 @@ impl LocalStakingApiHelper {
 
     pub fn receive_stake(
         &self,
-        // address of the user who originally called stake_remote
```
ethanfrey, line 48: Good fixes

contracts/native-staking/src/multitest.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    assert_eq!(
+        staking.proxy_by_owner(user1.to_owned()).unwrap(),
+        ProxyByOwnerResponse {
+            proxy: "contract1".to_string(),
+        }
+    );
+
+    assert_eq!(
+        staking.owner_by_proxy("contract1".to_string()).unwrap(),
```
ethanfrey, line 86: Good checks here.
One nit. I wouldn't hardcode the proxy address (in 81), but pull out a variable, asset not empty, and then do reverse query on the same variable

contracts/native-staking/src/multitest.rs
```diff
@@ -0,0 +1,270 @@
...
+    .unwrap();
+    staking
+        .local_staking_api_proxy()
+        .receive_stake(user1.to_owned(), stake_msg)
+        .with_funds(&coins(50, OSMO))
+        .call(owner) // called from vault
+        .unwrap();
+
+    // Check that same proxy is used
+    assert_eq!(
```
ethanfrey, line 111: And use above assertions here.

contracts/native-staking/src/multitest.rs
```diff
@@ -0,0 +1,270 @@
...
+    })
+    .unwrap();
+    staking
+        .local_staking_api_proxy()
+        .receive_stake(user1.to_owned(), stake_msg)
+        .with_funds(&coins(100, OSMO))
+        .call(owner) // called from vault
+        .unwrap();
+
+    assert_eq!(
```
ethanfrey, line 78: nb: if we are asserting this, maybe assert there is nothing registered before first stake?

contracts/native-staking/src/multitest.rs
```diff
@@ -0,0 +1,270 @@
...
+            coin(100, OSMO),
+            to_binary(&msg::StakeMsg {
+                validator: validator.to_owned(),
+            })
+            .unwrap(),
+        )
+        .call(user)
+        .unwrap();
+
+    // Vault has half the funds
```
ethanfrey, line 245: I would add one more check to this test case - that the user has a lien at this point (of 100 OSMO) and has no lien at the end. (Which asserts we actually call proper method, not just sending it back)

ethanfrey (approved):
Looks good.
Some minor comments on how to refine the test cases a bit.

contracts/native-staking/src/contract.rs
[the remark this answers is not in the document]
maurolacy: Gave it a try, but `into_result` complains about missing from impls.

contracts/native-staking/src/multitest.rs
maurolacy: That's a test that technically belongs to the `vault` test suite. But yes, wouldn't hurt to add it here as well.

contracts/native-staking/src/multitest.rs
maurolacy: There's no query to do that at the moment.

maurolacy: There's no query to do that at the moment. But, I can check that there's no proxy assigned to that user.

Issue #12, opened 2023-04-24
Allow staking on multiple remote validators


ethanfrey:
If a user has eg 100 tokens on a remote staking contract, they should be able to send 20 to one validator, 30 to another, and 50 to a third. And get the proper rewards and slashing, just as if they were 3 different users

ethanfrey:
This was closed by #39

PR #39, opened 2023-05-31
Picking the validator for stakes in external staking

hashedone:
Follow up of #37 

Addresses #13

PR made to `remote-staking` branch to hide any changes before it - it will be rebased and undrafted when #37 is merged. Please do not review it before #37 - it is built on top of it.

Implements only the second task mentioned in https://github.com/osmosis-labs/mesh-security/issues/13#issuecomment-1569538835.

Next - tests.

The most unsure thing here is if there is a need to keep track of which validators stake we are releasing - I assumed that it doesn't matter, and after calling `unbound` on tokens, they just are being unbound to the user.

If there were a need to distinguish it in the future, it would be very simple - proper fields will be moved from `Stake` to `User`.

A relevant question for the future is - are tokens rebounding period relevant in terms of distributing rewards (so if they are automatically "unbound" when `unbound` is called, just cannot be claimed before the unbound period, or if they are considered bound up until unbound period passes). I assumed the former because of implementation simplification. However, it will be easy to change it.

Finally, validator handling - the validator should always be sent for consistency of API. I used `String` instead of `Addr` to keep the validator address, as I assume it is an address on another chain and might be invalid on a local chain - let me know if it should be changed. To "query" the validator for a user, one can call the `stakes` query; for now, he will always get one query back. I didn't add any query for the only validator, as it seems redundant and possibly deprecated via #12.

If we want to have a validator being optional, then there should probably be the notion of a "default" validator per user (for now - the only validator; for the future - a special one that the user should be able to update), and the per contract (so we use it if the user does not have default validator and didn't provide one) - probably configurable on instantiation.


ethanfrey:
> The most unsure thing here is if there is a need to keep track of which validators stake we are releasing - I assumed that it doesn't matter, and after calling unbound on tokens, they just are being unbound to the user.

When unbonding, it is important to know which validator is being unbonded (to communicate this with the consumer).
When releasing to the vault, it doesn't matter (just stake).

When slashing is implemented, we will need to be able to slash pending_unbonding based on validator, so you could store that info somewhere,  but then again, there will probably be more changes needed by then and better just to keep it simple and solve the current use case.

> A relevant question for the future is - are tokens rebounding period relevant in terms of distributing rewards (so if they are automatically "unbound" when unbound is called, just cannot be claimed before the unbound period, or if they are considered bound up until unbound period passes). I assumed the former because of implementation simplification. However, it will be easy to change it.

Unbonding tokens do not collect any rewards. Just like normal PoS staking

> Finally, validator handling - the validator should always be sent for consistency of API. I used String instead of Addr to keep the validator address, as I assume it is an address on another chain and might be invalid on a local chain - let me know if it should be changed. To "query" the validator for a user, one can call the stakes query; for now, he will always get one query back. I didn't add any query for the only validator, as it seems redundant and possibly deprecated via https://github.com/osmosis-labs/mesh-security/issues/12.

Use `String`. Validator addresses are not `Addr` in any case, even on local chain.
When adding IBC, we will need to add more Validator info, asset conditions, sync state with consumer, hold tendermint pubkey for possible slashing, etc.

> If we want to have a validator being optional, then there should probably be the notion of a "default" validator per user (for now - the only validator; for the future - a special one that the user should be able to update), and the per contract (so we use it if the user does not have default validator and didn't provide one) - probably configurable on instantiation.

For MVP we can have each user with only one validator. And maybe some way of changing this (rebonding?), but with all their stake set on one remote validator per chain. By v2, they should be able to distribute among multiple validators. If it is easy to handle this now, go for it. If not, just use one validator per user for now.

contracts/external-staking/src/cross_staking_api.rs
```diff
@@ -32,14 +33,35 @@ impl CrossStakingApi for ExternalStakingContract<'_> {
...
+        // proper structure keeping `(user, validator)` as bound index
 
-        self.users.save(ctx.deps.storage, &owner, &user)?;
+        let stakes: Vec<_> = self
+            .stakes
+            .prefix(&owner)
+            .range(ctx.deps.storage, None, None, Order::Ascending)
+            .collect();
+
+        ensure_eq!(stakes.len(), 1, ContractError::InvariantsNotMet);
```
ethanfrey, line 47: This whole thing would be simpler code-wise to let a user bond to multiple validators.

Let's just remove that invairant. What other code would break on it?
* Unbond would require specifying which validator (which is a good idea anyway)
* We add the validator info to the pending unbonds (commented below in state)

I would just make full multi-validator staking as it seems not much more complexity here

contracts/external-staking/src/contract.rs
```diff
@@ -170,20 +187,27 @@ impl ExternalStakingContract<'_> {
...
 
-    /// Queries for user-related info
+    /// Queries for stake info
     ///
-    /// If user not existin in the system is queried, the default "nothing staken" user
-    /// is returned - also if user is not a valid address.
+    /// If stake is not existing in the system is queried, the default "nothing staken" is returned
     #[msg(query)]
-    pub fn user(&self, ctx: QueryCtx, user: String) -> Result<User, ContractError> {
+    pub fn user(
```
ethanfrey, line 194: Why called user here?
If you want a singular `stake` to go along with `stakes` below, I am all down with that.
We should maintain the user query, which shows pending unbonds

contracts/external-staking/src/contract.rs
```diff
@@ -66,7 +68,12 @@ impl ExternalStakingContract<'_> {
     /// Schedules tokens for release, adding them to the pending unbounds. After `unbounding_period`
     /// passes, funds are ready to be released with `claim` call by the user
     #[msg(exec)]
-    pub fn unstake(&self, ctx: ExecCtx, amount: Coin) -> Result<Response, ContractError> {
+    pub fn unstake(
+        &self,
+        ctx: ExecCtx,
+        validator: String,
```
ethanfrey, line 74: You have added this here already, let's just go full multi-validator staking. You have done 80% of the work and it is more trouble to enforce one validator than just embrace multiple (and where we want to go in the future anyway)

contracts/external-staking/src/contract.rs
```diff
@@ -75,17 +82,25 @@ impl ExternalStakingContract<'_> {
...
+        self.stakes
+            .save(ctx.deps.storage, (&ctx.info.sender, &validator), &stake)?;
+
+        let mut user = self
+            .users
+            .may_load(ctx.deps.storage, &ctx.info.sender)?
+            .unwrap_or_default();
 
         let release_at = config.unbounding_period.after(&ctx.env.block);
         let unbound = PendingUnbound {
```
ethanfrey, line 106: As commented below, just store the validator info here on each PendingUnbond (for future slashing) and we got full multiple validator cross-staking :+1:

contracts/external-staking/src/contract.rs
```diff
@@ -24,6 +24,7 @@ fn clamp_page_limit(limit: Option<u32>) -> usize {
 
 pub struct ExternalStakingContract<'a> {
     pub config: Item<'a, Config>,
+    pub stakes: Map<'a, (&'a Addr, &'a str), Stake>,
```
ethanfrey, line 27: Good to comment here what the args are `(user, validator)` as it is not trivially obvious without reading more code

ethanfrey (commented):
Nice start. Some things would improve from refactors on #37 

Otherwise, it seems like you did much multi-validator work already and it would be good to just go full multi-validator cross-staking already

contracts/external-staking/src/state.rs
```diff
@@ -13,18 +13,14 @@ pub struct Config {
...
+pub struct Stake {
     /// How much tokens user staken and not in unbonding period
     /// via this contract
     pub stake: Uint128,
-    /// List of token batches scheduled for unboding
-    ///
-    /// Items should only be added to the end of this list, with `release_at` being
-    /// `unbonding_period` after current time - this way this is guaranteed to be
-    /// always sorted (as time is guaranteed to be monotonic).
+    /// List of token batches scheduled for unbouding
```
ethanfrey, line 23: The problem of doing this per-validator is the overhead of generally "unbonding all" for a user.
I guess it is O(N) of the number of validators they have bonded to. And they control that, so just upping their gas cost. Not a blocker, just surprising. I will look at code

contracts/external-staking/src/contract.rs
hashedone: True, forgot to do that

contracts/external-staking/src/contract.rs
hashedone: Let's do it in another PR - less to test, less to merge. I am fine delivering it.

contracts/external-staking/src/contract.rs
```diff
@@ -169,23 +199,153 @@ impl ExternalStakingContract<'_> {
...
+            );
+
+            let owner = ctx.deps.api.addr_validate(&owner)?;
+
+            let msg: ReceiveVirtualStake = from_binary(&msg)?;
+
+            // For now we assume there is only single validator per user, however we want to maintain
+            // proper structure keeping `(user, validator)` as bound index
+
+            let stakes: Vec<_> = self
```
ethanfrey, line 303: Let's remove this and support multi (all the other logic does).

```rust
let key = (&owner, &msg.validator);
let mut stake = self.stakes.may_load(ctx.deps.storage, key)?.unwrap_or_default();
stake.stake += amount.amount;
self.stakes.save(ctx.deps.storage, key, &stake)?;
```

A lot fewer lines!

contracts/external-staking/src/contract.rs
```diff
@@ -115,13 +127,24 @@ impl ExternalStakingContract<'_> {
...
 
-        let released = user.release_pending(&ctx.env.block);
-        self.users.save(ctx.deps.storage, &ctx.info.sender, &user)?;
+        let released: Uint128 = stakes
+            .into_iter()
+            .map(|(validator, mut stake)| {
+                let released = stake.release_pending(&ctx.env.block);
+                self.stakes
+                    .save(ctx.deps.storage, (&ctx.info.sender, &validator), &stake)
+                    .map(|_| released)
```
ethanfrey, line 143: Why not:

```rust
self.stakes.save(ctx.deps.storage, (&ctx.info.sender, &validator), &stake)?;
Ok(released);
```

Seems clearer than the map here (though I see you loving functional style)

contracts/external-staking/src/contract.rs
```diff
@@ -115,13 +127,24 @@ impl ExternalStakingContract<'_> {
...
+            })
+            .fold(Ok(Uint128::zero()), |acc, released| {
```
ethanfrey, line 145: We really need to `impl Sum for Result<Uint128, E>` (but in cosmwasm-std and probably not possible)

contracts/external-staking/src/contract.rs
```diff
@@ -115,13 +127,24 @@ impl ExternalStakingContract<'_> {
...
+                let acc = acc?;
+                released.map(|released| released + acc)
+            })?;
 
         let release_msg = config.vault.release_cross_stake(
```
ethanfrey, line 150: Before sending tokens, check if `released.is_zero()` and exit early without sending tokens if so.
Otherwise you will get some weird error message about sending 0 tokens.

If you want to error, make some explicit `ContractError::NothingToWithdraw` to return here

contracts/external-staking/src/contract.rs
```diff
@@ -202,15 +226,49 @@ impl ExternalStakingContract<'_> {
...
+    #[msg(query)]
+    pub fn stakes(
+        &self,
+        ctx: QueryCtx,
+        user: String,
+        start_after: Option<String>,
+        limit: Option<u32>,
+    ) -> Result<StakesResponse, ContractError> {
+        let limit = clamp_page_limit(limit);
+        let user = Addr::unchecked(user);
```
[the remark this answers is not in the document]
ethanfrey: Pinging this comment

contracts/external-staking/src/contract.rs
```diff
@@ -169,23 +199,153 @@ impl ExternalStakingContract<'_> {
         let limit = clamp_page_limit(limit);
 
         let start_after = start_after.map(Addr::unchecked);
-        let bound = start_after.as_ref().and_then(Bounder::exclusive_bound);
+        let bound = start_after.as_ref().map(PrefixBound::exclusive);
```
ethanfrey, line 203: This is getting to be a super complex method of limited value. (While stakes has clear value).

Maybe remove it?
Or just comment it out until the frontend guys ask for it?

I think it will take some time to test to make sure it works with many users and stakes and all, and not that important.

contracts/external-staking/src/contract.rs
ethanfrey: But in general, the logic in this method looks good

contracts/external-staking/src/contract.rs
```diff
@@ -169,23 +200,153 @@ impl ExternalStakingContract<'_> {
...
+        let stakes = self
+            .stakes
+            .prefix(&user)
             .range(ctx.deps.storage, bound, None, Order::Ascending)
             .map(|item| {
-                item.map(|(addr, user)| UserInfo {
-                    addr: addr.into(),
-                    stake: user.stake,
-                    pending_unbonds: user.pending_unbonds,
+                item.map(|(validator, stake)| StakeInfo {
```
ethanfrey, line 256: 👍

contracts/external-staking/src/state.rs
ethanfrey: The code around unbonding and withdrawing makes sense. Nice design. let's keep it this way.

ethanfrey: ps. it is `unbond` not `unbod` or `unboud`

contracts/external-staking/src/contract.rs
ethanfrey: You could also optimize to only save if something was released (and skip those where nothing was ready). Much easier without the map ;)

```rust
if !released.is_zero() {
  self.stakes.save(ctx.deps.storage, (&ctx.info.sender, &validator), &stake)?;
}
Ok(released);
```

ethanfrey (approved):
Looks good.
Mostly style issues.
Only real bug I found is cleanly handling withdraw_unbond with 0 tokens released. And an optimization when claiming.

I would really not spend more time on the users list query. Leave the code (as comment) and it can be properly tested and finished if needed later

contracts/external-staking/src/state.rs
```diff
@@ -21,6 +21,10 @@ pub struct Stake {
     /// via this contract
     pub stake: Uint128,
     /// List of token batches scheduled for unbouding
+    ///
+    /// Items should only be added to the end of this list, with `release_at` being
+    /// `unbonding_period` after current time - this way this is guaranteed to be
+    /// always sorted (as time is guaranteed to be monotonic).
```
ethanfrey: nice comment

contracts/external-staking/src/contract.rs
hashedone: In general, I am thinking about including `itertools` at some point - it is not this huge, and it contains all sort of functions for working on iterators over results - eg, https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.fold_ok which is still a fold, but `fold_ok(Uint128::zero(), Add::add)` reads far more clear.

contracts/external-staking/src/contract.rs
hashedone: Nah, no reason for the error. Also - there are also no tokens sent, only messages triggered, nothing should happen. However, it is a fact that there is no reason to send msg in such a case, will fix.

contracts/external-staking/src/contract.rs
hashedone: In general, I am thinking about including `itertools` at some point - it is not this huge, and it contains all sort of functions for working on iterators over results - eg, https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.fold_ok which is still a fold, but `fold_ok(Uint128::zero(), Add::add)` reads far more clear.

contracts/external-staking/src/contract.rs
hashedone: Nah, no reason for the error. Also - there are also no tokens sent, only messages triggered. Nothing should happen. However, it is a fact that there is no reason to send msg in such a case, will fix it.

contracts/external-staking/src/contract.rs
hashedone: Ok, I'll remove it, not a problem

contracts/external-staking/src/contract.rs
hashedone: Yea, it would be far easier.

Actually, I was kind of sure multi-validators are not supported because some IBC-thing is not yet defined, but honestly - for this contract, it is super easy to implement.

contracts/external-staking/src/state.rs
hashedone: Ahh I already grep for `unbound` to fix it, but another typo here which rg missed :D

I am not super happy about O(n) as well, but it is always O(n) dependent on the amount of unbonds performed. I don't see a way to go around that.

PR #40, opened 2023-06-01
Various fixes to Virtual Staking contract

ethanfrey:
Virtual staking is being tested in system tests.

Errored on "Not Found" due to uninitialized Items.
This fixes two cases this could occur.

Also, use BTreeMap not HashMap due to consistency concerns

Finally, use the `VirtualStakingMsg::{Bond, Unbond}` messages rather than `StakingMsg::{Delegate, Undelegate}`

This is being tested in https://github.com/osmosis-labs/mesh-security-sdk/pull/22

### DIFF
--- a/contracts/consumer/virtual-staking/src/contract.rs
+++ b/contracts/consumer/virtual-staking/src/contract.rs
@@ -8,7 +8,9 @@ use cosmwasm_std::{
 use cw2::set_contract_version;
 use cw_storage_plus::{Item, Map};
 use cw_utils::nonpayable;
-use mesh_bindings::{TokenQuerier, VirtualStakeCustomMsg, VirtualStakeCustomQuery};
+use mesh_bindings::{
+    TokenQuerier, VirtualStakeCustomMsg, VirtualStakeCustomQuery, VirtualStakeMsg,
+};
 use sylvia::types::{ExecCtx, InstantiateCtx, QueryCtx};
 use sylvia::{contract, schemars};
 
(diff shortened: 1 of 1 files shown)


contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -137,7 +141,7 @@ fn calculate_rebalance(
     desired: Vec<(String, Uint128)>,
     denom: &str,
 ) -> Vec<CosmosMsg<VirtualStakeCustomMsg>> {
-    let mut desired: HashMap<_, _> = desired.into_iter().collect();
+    let mut desired: BTreeMap<_, _> = desired.into_iter().collect();
```
hashedone, line 144: Why this change? Did you profile it? It is very uncommon for btreemap to be more efficient, as it is not cache friendly, and if it is in linear memory, then it is performing linear search... If btreemap is more efficient in hashmap, then it means that the problem is hashing, and probably `Vec<(_, _)>` would be even faster, did you consider changing the hasher (eg https://crates.io/crates/rustc-hash)

hashedone (approved):
LGTM, but I am very concerned about the map change - I would like to know what is the root reason for that. I assume there is one so approve, but I am really curios about that.

ethanfrey:
There was a discussion (mainly between Alex and Mauro) about non-determinism of HashMaps. Which is totally true in Go.

I said not so in Rust, but Mauro said it was unstable between runs, and I figured why not change.

I used HashMap for performance. If you want to rewrite it, while keeping a predictable ordering of the messages, I would be happy to review

PR #42, opened 2023-06-01
Tests for external staking

### LINKED ISSUE #13: Implement Remote Staking contract (no rewards)

hashedone:
Follow up of #39 

Closes #13 

Implements only the third task mentioned in https://github.com/osmosis-labs/mesh-security/issues/13#issuecomment-1569538835.

Next - distribution.

Multitests + minor bugfixes found by tests.


contracts/vault/src/contract.rs
```diff
@@ -263,6 +263,22 @@ impl VaultContract<'_> {
         Ok(resp)
     }
 
+    /// Returns a singl claim between the user and lienholder
```
ethanfrey, line 266: singl -> single

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+
+use crate::contract::cross_staking::test_utils::CrossStakingApi;
+use crate::contract::multitest_utils::{CodeId, ExternalStakingContractProxy};
+use crate::error::ContractError;
+use crate::msg::{ReceiveVirtualStake, StakeInfo};
+use crate::state::Stake;
+
+use anyhow::Result as AnyResult;
+
+const OSMO: &str = "OSMO";
```
ethanfrey, line 19: Nit: if you want to be closer to reality, use `uosmo` here (OSMO only exists in UIs not on the chain)

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+
+    let max_slash = contract.cross_staking_api_proxy().max_slash().unwrap();
+    assert_eq!(max_slash.max_slash, Decimal::percent(5));
+}
+
+#[test]
+fn staking() {
+    let users = ["user1", "user2"];
+
+    let app = MtApp::new(|router, _api, storage| {
```
ethanfrey, line 81: Minor note for Sylvia:
It would be nice to add some helper for the simple setup of just setting balances (and only use closure for more important cases). Like keep `MtApp:::new` as is, but something like this, which takes `balances: &[(&str, Vec<Coin>])`

```rust
let app = MtApp::with_balances(&[
  (users[0], coins(300, OSMO)),
  (users[1], coins(300, OSMO))
];
```

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+            coin(200, OSMO),
+            to_binary(&ReceiveVirtualStake {
+                validator: validators[1].to_string(),
+            })
+            .unwrap(),
+        )
+        .call(users[1])
+        .unwrap();
+
+    vault
```
ethanfrey, line 165: Great stuff.

For completeness, I would add one more stake, and ensure it returns an error (no more collateral left for lien)

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+            coin(300, OSMO),
+            to_binary(&ReceiveVirtualStake {
+                validator: validators[0].to_string(),
+            })
+            .unwrap(),
+        )
+        .call(users[1])
+        .unwrap();
+
+    // Properly unstake some tokens
```
ethanfrey, line 351: nice cases and error checking here

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+        .stake(users[1].to_string(), validators[0].to_string())
+        .unwrap();
+    assert_eq!(stake.stake.u128(), 240);
+
+    let stake = contract
+        .stake(users[1].to_string(), validators[1].to_string())
+        .unwrap();
+    assert_eq!(stake.stake.u128(), 0);
+
+    // Another timetravel - just enough for first bathc of stakes to release,
```
ethanfrey, line 490: Nice setup for the withdrawing process and ensuring this works with multiple batches pending.

Small typo: bathc -> batch

ethanfrey (approved):
Great stuff.

Left a few comments on typos. Please fix and merge

contracts/external-staking/src/state.rs
```diff
@@ -49,13 +49,13 @@ impl Stake {
         };
 
         // First item is still not ready for release
-        if self.pending_unbonds[0].release_at < info.time {
+        if self.pending_unbonds[0].release_at > info.time {
```
maurolacy, line 52: That's what tests are for.

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+    // to early for second batch
```
maurolacy, line 491: to -> too

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+    app.app_mut().update_block(|block| {
+        block.height += 1;
+        block.time = block.time.plus_seconds(50);
+    });
+
+    // Withdrawing liens
+    contract.withdraw_unbonded().call(users[0]).unwrap();
+    contract.withdraw_unbonded().call(users[1]).unwrap();
+
+    // Now claims on vauld got reduced, but only for first batch amount
```
maurolacy, line 501: vauld -> vault.

contracts/external-staking/src/multitest.rs
```diff
@@ -0,0 +1,543 @@
...
+        .claim(users[0].to_owned(), contract.contract_addr.to_string())
+        .unwrap();
+    assert_eq!(claim.amount.u128(), 250);
+
+    let claim = vault
+        .claim(users[1].to_owned(), contract.contract_addr.to_string())
+        .unwrap();
+    assert_eq!(claim.amount.u128(), 240);
+
+    // Moving forward more, passign through second bath pending duration
```
maurolacy, line 512: passign -> passing

maurolacy (approved):
LGTM.

contracts/external-staking/Cargo.toml
```diff
@@ -36,6 +36,9 @@ thiserror        = { workspace = true }
 sylvia        = { workspace = true, features = ["mt"] }
 cw-multi-test = { workspace = true }
 anyhow        = { workspace = true }
+mesh-vault = { workspace = true, features = ["mt"] }
+mesh-native-staking-proxy = { workspace = true, features = ["mt"] }
```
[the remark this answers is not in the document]
hashedone: Only vault initialization. I was looking for a way to fake it somehow, but I need to pass the proper message and so on, so I decided to add them and just instantiate - doesn't hurt for tests.

contracts/external-staking/src/multitest.rs
hashedone: It completely does not matter, but will change just for consistency.

contracts/external-staking/src/multitest.rs
hashedone: In general, what would happen for sure:
1. `App::new()` would be changed to match the `App::new(...)` from multitest - will take instantiation closure
2. `AppBuilder` would be created to mimic the one from MT - allowing setting up custom modules
3. There would be an additional layer between the builder and the app - `AppConfig` which would have some additional functions knowing what modules are used. Eg. when default `BankKeeper` is used it would have function like `initial_balances(...)` - exactly what you said. It would be another builder state, so when there is another config option it doesn't add a dozen of functions passing it in different combinations. Obviously `Builder` would still be able to return an `App` directly if we don't want to config it.
4. There would be added `From<cw_multi_test::App> fo sylvia::App` implementation, so it would still be possible to create mt app manually and then convert to sylvia one - in case some MT feature would be added, and we will not update sylvia automatically.

contracts/external-staking/src/state.rs
hashedone: Yup, that is why I really didn't check it 100 times before, as it would be burned time - I knew I am writing tests for it in next PR and it would be verified there.

PR #44, opened 2023-06-02
Try to build release workflow

ethanfrey:
Need an expert on GitHub actions.
I want to port this logic over: https://github.com/CosmWasm/cw-plus/blob/main/.circleci/config.yml#L536-L626
And am using https://trstringer.com/github-actions-create-release-upload-artifacts/ as a guide.

I can't figure out how to loop over all contracts, so right now, I just try to build one of them and see if I can get that to work, so hopefully someone can expand it

Update: that article recommended an obsolete action. I am trying out https://github.com/softprops/action-gh-release which seems maintained

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,9 +13,6 @@ jobs:
   release:
     name: Build Release Artifacts
     runs-on: ubuntu-latest
-    defaults:
-      # hack to see if this works at all, we need to do all contracts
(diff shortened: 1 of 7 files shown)


ethanfrey:
It builds now, yay!

See https://github.com/osmosis-labs/mesh-security/releases/tag/v0.2.0-alpha.3 as an example

PR #45, opened 2023-06-02
External staking rewards distribution

hashedone:
Follow up of #42 

Addresses #13 

Implements only the fourth task mentioned in https://github.com/osmosis-labs/mesh-security/issues/13#issuecomment-1569538835.

I miss `Int256` and `Int128` really badly - we need to finish math. Not proud of a workaround, but at least I didn't copy-paste the whole type from some other contract. And it should work. I am aware that due to `u2` I could just `wrapping_add` and `wrapping_sub`, but then we lose debug control of out of bound.

Next upcoming in the next PR.

### DIFF
--- a/contracts/external-staking/src/contract.rs
+++ b/contracts/external-staking/src/contract.rs
@@ -1,9 +1,12 @@
+use std::cmp::Ordering;
+
 use cosmwasm_std::{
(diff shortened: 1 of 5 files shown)


hashedone:
Closed in favor of #46

PR #49, opened 2023-06-06
Define IBC communication

### LINKED ISSUE #7: Finalize documentation

ethanfrey:
Closes #7 

- [ ] Document different communication channels (control, rewards, slashing)
- [ ] Define IBC control message types
- [ ] Define approach for reward payments
- [ ] Rough description of how to handle slashing

### DIFF
--- a/packages/apis/src/ibc.rs
+++ b/packages/apis/src/ibc.rs
@@ -1,0 +1,74 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::Uint128;
+
+/// These are messages sent from provider -> consumer
+/// ibc_packet_receive in converter must handle them all.
+/// Each one has a different ack to be used in the reply.
+#[cw_serde]
+pub enum ProviderMsg {
+    /// This should be called on initialization to get current list of validators.
+    /// Any changes to the set should be sent as a ConsumerMsg::UpdateValidatorSet
+    ListValidators {},
+    /// This should be called when we lock more tokens to virtually stake on a given validator
+    Stake {
+        validator: String,
+        /// We can't really use Coin here, as which denom? Remote staking denom, or how that
+        /// token appears as ibc denom. Just encode the amount.
+        amount: Uint128,
+    },
+    /// This should be called when we begin the unbonding period of some more tokens previously virtually staked
+    Unstake {
+        validator: String,
+        /// We can't really use Coin here, as which denom? Remote staking denom, or how that
+        /// token appears as ibc denom. Just encode the amount.
+        amount: Uint128,
+    },
+}
+
+/// Ack sent for ProviderMsg::ListValidators
+#[cw_serde]
+pub struct ListValidatorsAck {
+    pub validators: Vec<Validator>,
+}
+
+/// Ack sent for ProviderMsg::Stake
+#[cw_serde]
+pub struct StakeAck {}
+
+/// Ack sent for ProviderMsg::Unstake
+#[cw_serde]
+pub struct UnstakeAck {}
+
+#[cw_serde]
+pub struct Validator {
+    /// This is the validator address used for delegations and rewards
+    pub valoper_address: String,
+
+    // TODO: is there a better type for this? what encoding is used
+    /// This is the *Tendermint* public key, used for signing blocks.
+    /// This is needed to detect slashing conditions
+    pub pub_key: String,
+
+    /// This is the moniker of the validator, used for display purposes
+    pub moniker: String,
+}
+
+/// These are messages sent from consumer -> provider
+/// ibc_packet_receive in external-staking must handle them all.
+#[cw_serde]
+pub enum ConsumerMsg {
+    /// This should be sent when the validator set changes.
+    /// Add includes full validator info to add.
+    /// Remove includes only the valoper_address to remove.
+    ///
+    /// Question: This should not be sent when a validator enters/leaves the "active set",
+    /// but rather when they are tombstoned.
+    UpdateValidatorSet {
+        add: Vec<Validator>,
+        remove: Vec<String>,
+    },
+}
+
+/// Ack sent for ConsumerMsg::UpdateValidatorSet
+#[cw_serde]
+pub struct UpdateValidatorSetAck {}
--- a/packages/apis/src/lib.rs
+++ b/packages/apis/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod converter_api;
 pub mod cross_staking_api;
+pub mod ibc;
 pub mod local_staking_api;
 pub mod price_feed_api;
 pub mod vault_api;


docs/ibc/Staking.md
```diff
@@ -1,3 +1,136 @@
...
+`external-staking` contract should immediately call back to the vault to release
+the lein for the amount of that packet. If a success ack comes back, it should
+actually apply the staking changes locally.
+
+Likewise, when unstaking, we first create an `Unstake` packet. However, here 
+we want to avoid the situation of `Unstaking` 100% of the tokens multiple times,
+as this would apply invalid state changes to the consumer. We need to enforce this 
+limit on the provider side (which is handled by the vault in the staking case).
+
+For `Unstake`, we should update a local "unstaking" value on the `(user, validator)`
```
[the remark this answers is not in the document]
ethanfrey: Not needed there. We need it for unstake to avoid going below 0, as we track how much can be unstaked in this contract.

For staking, the assurance that they haven't staked more than they have collateral is enforced in the vault. This is out "pending_staked" in a way. The change *is* written to vault before we try to send IBC message, which is why it is important to revert that by releasing the lien if the packet fails.

I don't see the need for tracking this a second time in the external-stake

docs/ibc/Staking.md
```diff
@@ -1,3 +1,136 @@
...
+as well as the total amounts staked. 
+
+If we reorder the diffs, it is possible to get a different result, so we need to be careful
+about relying on Unordered channels. Imagine Stake 100 and Unstake 50. If Unstake goes first,
+it would return an error, yet the Stake would apply properly, leaving a total of 100 not 50.
+Furthermore, if a packet is dropped, and the diff is not applied, the two sides will
+get out of sync, where eg the Provider believes there is 500k staked to a given validator,
+while the Consumer believes there is 400k. 
+
+At the same time, Ordered channels are fragile, in that a single timed-out or undeliverable packet
```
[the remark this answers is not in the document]
ethanfrey: Yup.

CRDTs are not possible as they say every ordering should give the same final state. Which makes enforcing invariants like "never go below 0" impossible unless it is a monotonically increasing function.

If +10 -5  is valid, then a CRDT would require -5 +10 to also be valid and have the same meaning.
Yet this would lead to a negative stake error. 

I was really into CRDTs for better data syncing (between devices with spotty coverage) in my last job before blockchain. But it doesn't really work in blockchain, as it really asserts ordering and "always >=  0" invariants to avoid "double spend attack".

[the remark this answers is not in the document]
ethanfrey: We can be inspired by vector clocks, but again, there is no merge function here. There is one state.

We can imagine all possible paths / orderings and then ensure that they all lead to consistent states, but possibly different ones.

docs/ibc/Staking.md
```diff
@@ -1,3 +1,136 @@
...
+actually apply the staking changes locally.
+
+Likewise, when unstaking, we first create an `Unstake` packet. However, here 
+we want to avoid the situation of `Unstaking` 100% of the tokens multiple times,
+as this would apply invalid state changes to the consumer. We need to enforce this 
+limit on the provider side (which is handled by the vault in the staking case).
+
+For `Unstake`, we should update a local "unstaking" value on the `(user, validator)`
+staking info, but not create a claim nor apply a diff to the validator. 
+We ensure that this "unstaking" amount can never be larger than the properly staked
```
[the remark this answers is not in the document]
ethanfrey: Can you expand on this point?

packages/apis/src/ibc.rs
```diff
@@ -0,0 +1,73 @@
...
+        /// of the price feed.
+        unstake: Coin,
+    },
+}
+
+/// Ack sent for ProviderMsg::ListValidators
+#[cw_serde]
+pub struct ListValidatorsAck {
+    pub validators: Vec<Validator>,
+}
```
JakeHartnell, line 34: Is there ever a world in which we have to paginate this? A chain with 300 validators is technically possible...

JakeHartnell: Almost would rather not worry about this ATM though.

packages/apis/src/ibc.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub valoper_address: String,
+
+    // TODO: is there a better type for this? what encoding is used
+    /// This is the *Tendermint* public key, used for signing blocks.
+    /// This is needed to detect slashing conditions
+    pub pub_key: String,
+
+    // TODO: remove this? add other metadata?
+    /// This is the moniker of the validator, used for display purposes
+    pub moniker: String,
```
JakeHartnell, line 56: I would remove. With the validator address frontends or indexers can fetch this data, and I can't imagine we would use moniker in the contract.

docs/ibc/Staking.md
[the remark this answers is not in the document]
alpe: Consumer and. provider side can become out of sync on the staked amount when slashing happens.
Slashing is originated on the consumer side and reduces the effectively staked amount.  It will take some time to propagate the event to the provider side. During this period both sides are out of sync and the real stake amount is < provider side stake amount.

docs/ibc/Staking.md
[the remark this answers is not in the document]
alpe: Fair enough, when this is tracked in the vault. The user should be prohibited to stake the same tokens twice while it is in-flight.

docs/ibc/Staking.md
```diff
@@ -1,3 +1,136 @@
...
+
+We could invent some "re-sync" mechanism, but would have to be careful mixing this with
+in-flight messages. For example, the provider sends a `Stake` message to the consumer,
+which returns an error for whatever reason. Before the ack has arrived, the provider
+sends a re-sync message with the entire content of it's local state.
+
+Upon receiving that "re-sync" message, the consumer updates all tables and triggers the
+appropriate virtual staking commands. However, the error ack for the `Stake` packet
+now lands on the provider and it "undoes" the Stake action. This will be applied
+on top of the snapshot that it sent to the consumer, thus modifying it and the two sides
```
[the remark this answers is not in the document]
alpe: I got a bit lost in technical details. I think if we can switch from push (consumer) to poll (provider side) as start of the process, things should just work on an ordered channel. The provider would ask for a state sync on validator(s) and then receive the result. Any other operation started by the provider side should be completed before the sync arrives, except for slashing.

packages/apis/src/ibc.rs
```diff
@@ -0,0 +1,73 @@
...
+use cosmwasm_std::Coin;
+
+/// These are messages sent from provider -> consumer
+/// ibc_packet_receive in converter must handle them all.
+/// Each one has a different ack to be used in the reply.
+#[cw_serde]
+pub enum ProviderMsg {
+    /// This should be called on initialization to get current list of validators.
+    /// Any changes to the set should be sent as a ConsumerMsg::UpdateValidatorSet
+    ListValidators {},
```
alpe, line 11: I was wondering if we need to fetch and store a copy of the whole valset on the provider chain. With an optimistic approach, we would only store the validators that received delegates from the users (that were acked by the consumer side). That would reduce complexity in the system.
The frontend would need to read the full valset from the consumer chain directly. But this part is not clear to me. 

I wrote a bit about validation in:  https://github.com/osmosis-labs/mesh-security-sdk/issues/28#issuecomment-1578108546

docs/ibc/Staking.md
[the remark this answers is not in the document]
ethanfrey: Yes, very good point. I will elaborate on how we enforce this

docs/ibc/Staking.md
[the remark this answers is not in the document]
ethanfrey: I spent some time last night reflecting more and realize the *validator set* is actually a perfect use of CRDT. 

We just have to treat `Add(A), Remove(A)` the same as `Remove(A), Add(A)`, meaning not error on removing a non-existant validator, but store this drop message for later.

packages/apis/src/ibc.rs
ethanfrey: > With an optimistic approach, we would only store the validators that received delegates from the users (that were acked by the consumer side). That would reduce complexity in the system.

You are right. We don't *need* the info except for ones with stake, so we can slash.

But after reading your issue on validation, I figured it was best to enforce the validator that would receive delegation before creating the packet. It's not dangerous to delegate to an invalid validator address, it will just return an error in ack. But the UI experience is much easier to show error in transaction (a few seconds after signing, when user is on the page), then error in ack (which may occur a minute or two later, maybe they closed the app already).

This is more for application design than enforcing correctness of the protocol.

docs/ibc/Staking.md
```diff
@@ -1,3 +1,178 @@
...
+they were originally sent.
+
+If there is a failure sending a Stake / Unstake action, the safest response is to
+undo the action locally and inform the user. We need solid UI here, as suddenly having
+a staking action disappear with no notification as to why will lead to serious confusion
+and cries of "bugs", when it was just an invisible, delayed error.
+
+An alternate approach here is to have some sort of "re-sync" design to re-synchronize the
+state on the two sides, but I argue this is too fragile in an asynchrnous environment.
+Much care should be taken to ensure that all applications errors are handles well.
```
maurolacy, line 134:
```suggestion
Much care must be taken to ensure that all application errors are handled well.
```

docs/ibc/Staking.md
```diff
@@ -1,3 +1,178 @@
...
+
+### Error handling
+
+When staking, create a `Stake` packet, but do not update the user's / validator's state
+in the `external-staking` contract. If an error ack comes back, then the
+`external-staking` contract should immediately call back to the vault to release
+the lein for the amount of that packet. If a success ack comes back, it should
```
maurolacy, line 141: lein -> lien

docs/ibc/Staking.md
```diff
@@ -1,3 +1,178 @@
...
+
+It is important that you **do not** use `#[cw_serde]` on the Rust struct as we explicitly
+want to **allow unknown fields**. This allows us to add new fields in the future.
+`#[cw_serde]` generates `#[serde(deny_unknown_fields)]` which would break this.
+
+Both sides must error if the protocol is not `mesh-security`. 
+
+The version is used to allow for future upgrades. The provider should always send the
+highest protocol version it supports to start the handshake. The consumer should
+error if the major version is higher than it's known versions (for now, anything besides `1`).
```
maurolacy, line 55: it's -> its

maurolacy (commented):
Good docs / comments. Just some syntax / style comments.

docs/ibc/Staking.md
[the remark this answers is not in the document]
ethanfrey: Oh, that is a very good point. We need a much longer section on slashing. But I will include a space to dig into this point more.

ethanfrey:
Okay, I cleaned up the smaller issues here, maybe good starting point now.

I will make a new PR on top of this to dig deeper into correctness of messages and unordered/CRDT items

docs/ibc/Overview.md
```diff
@@ -16,10 +16,10 @@ and discuss cross-chain slashing in another section.
 ```mermaid
 flowchart RL
   %%{init: {'theme': 'forest'}}%%
-
```
maurolacy, line 19: Nit: Remove trailing spaces, here and elsewhere.

docs/ibc/Overview.md
```diff
@@ -51,32 +51,32 @@ to authorize them:
...
 ## Protocol design
 
 We use this custom channel to communicate securely and directly
 between the two sides, sending messages about bonding and unbonding
-delegations. This is the principle form of communication,
-and we dig into the [Cross-Chain Staking Protocol](./Staking.md)
-in more depth.
+delegations, validator updates, and possibly more. This is the
+principle form of communication, called the
+[Control Channel](./ControlChannel.md), and is the core
```
maurolacy, line 71: 👍🏼  Good to separate this into its own file, and reference it.

docs/ibc/Overview.md
```diff
@@ -51,32 +51,32 @@ to authorize them:
...
+piece of the Mesh Security IBC protocol.
 
 We send reward tokens over a standard ics20 channel so that
 they are fungible with the native staking token currently sent
 between the chains (it will have the same denom as what's listed on the DEX).
 Besides moving the rewards, we need to inform the _External Staker_ that
 it has received rewards so it can distribute them to validators.
-We will do that either with a separate message in the control channel (above),
-or by making use of the ICS20 memo field to pass this info.
+We will do via ICA, or by making use of the ICS20 memo field to pass this info.
```
maurolacy, line 79:
```suggestion
We will do that via ICA, or by making use of the ICS20 memo field to pass this info.
```

docs/ibc/Validators.md
```diff
@@ -0,0 +1,205 @@
...
+
+The basic implementation without public key rotation is another expression of the same algorithm
+used in [`2P-Set`](<https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#2P-Set_(Two-Phase_Set)>). This is a proven CRDT, and if we implement it to match the spec, we have a
+guarantee of commutability.
+
+## Validator Key Rotation
+
+In addition to the basic operations described above, there is another operations we would like
+to support. This protocol is designed to handle Tendermint key rotation, such that a validator
+address may be associated with multiple public keys over the course of it's existence.
```
maurolacy, line 112: it's -> its

docs/ibc/Validators.md
```diff
@@ -0,0 +1,205 @@
...
+We wish to maintain a validator set `V` on the Provider with the following properties:
+
+- If no packet has been received for a given `x`, `x` is not in `V`
+- If `R(x)` has been received, `x` is not in `V`
+- If at least one `A(x, _, _)` has been received, but no `R(x)`, `x` is in `V` with:
+  - A set of all pubkeys, along with the block height they were first active from.  
+    (We may represent this as a sorted list without duplicates, but that is a mathematically
+    equivalent optimization)
+
+To ensure we can perform all this with the commutivity property, we look for a mapping
```
maurolacy, line 129: commutivity -> commutativity

docs/ibc/Validators.md
```diff
@@ -0,0 +1,205 @@
...
+    R(x) => x,
+    A(x, _) => x,
+};
+let old_state: Option<State> = VALIDATORS.may_load(storage, x)?;
+let new_state: State = match (old_state, op) {
+    (_, R(_)) => State::Tombstoned,
+    (Some(State::Tombstoned), _) => State::Tombstoned,
+    (None, A(_, p)) => Some(State::Active(p)),
+    // Ignore A if we have received state already
+    (Some(State::Active(p)), A(_, _)) => Some(State::Active(p)),
```
maurolacy, line 98: Remove `Some` from output?

If we ignore the subsequent `A`, what if they contain a different `p`?

docs/ibc/Validators.md
```diff
@@ -0,0 +1,205 @@
...
+    // This handles leaving the 2P-Set
+    (_, R(_)) => State::Tombstoned,
+    (Some(State::Tombstoned), _) => State::Tombstoned,
+
+    // Joining active set for the first time
+    (None, A(_, p, h)) => Some(State::Active([(p, h)])),
+    // Rotating the pubkey, adds to the set
+    (Some(State::Active(mut v)), A(_, p, h)) => {
+        // Add this transition
+        v.push((po, h));
```
maurolacy, line 182:
```suggestion
        v.push((p, h));
```

docs/ibc/Validators.md
```diff
@@ -0,0 +1,205 @@
...
+        // We sort the array from largest to smallest for quick compares on slashing
+        v.sort_by(|a, b| a.cmp(b).reverse());
+        // Remove all duplicates (handle if `K` was seen before)
+        v.dedup();
+        // Assert invariant violation if there are now multiple items same height
+        debug_assert_eq!(v.iter().filter(|(_, hx)| hx == &h).count(), 1);
+        Some(State::Active(updated, v)),
```
maurolacy, line 189:
```suggestion
        State::Active(v),
```

maurolacy: Why not use a `BTreeSet` instead of a `Vec`? And just check for existence before inserting a duplicate.

docs/ibc/Validators.md
```diff
@@ -0,0 +1,205 @@
...
+    (Some(State::Tombstoned), _) => State::Tombstoned,
+
+    // Joining active set for the first time
+    (None, A(_, p, h)) => Some(State::Active([(p, h)])),
+    // Rotating the pubkey, adds to the set
+    (Some(State::Active(mut v)), A(_, p, h)) => {
+        // Add this transition
+        v.push((po, h));
+        // We sort the array from largest to smallest for quick compares on slashing
+        v.sort_by(|a, b| a.cmp(b).reverse());
```
maurolacy, line 184: This comparison should be by height, not pub key. Can be done when implementing `Cmp` over the tuple / struct.
I think a `BTreeSet` will work fine here.

maurolacy (commented):
Partial review of `Overview` and `Validators`. Will continue during the day.

docs/ibc/ControlChannel.md
```diff
@@ -0,0 +1,118 @@
+# Control Channel
+
+sThis document describes the "control channel", which is a direct IBC channel between the
```
maurolacy, line 3: sThis -> This

docs/ibc/ControlChannel.md
```diff
@@ -0,0 +1,118 @@
...
+and any other metadata about the protocol (like validators).
+
+It is **not** used to [send reward tokens](./Rewards.md), which must be sent over
+the commonly accepted ICS20 interface, so they are fungible after receipt.
+
+It is also **not** used to [handle slashing](./Slashing.md), as there are concerns
+a malicious state machine would lie, so we demand original evidence of Tendermint
+misbehavior on the provider chain.
+
+This channel encompasses to sub-protocols, which are described in their own documents.
```
maurolacy, line 15: to -> two ?

docs/ibc/ControlChannel.md
```diff
@@ -0,0 +1,118 @@
...
+Unordered channels make it harder to prove guarantees for the application in an asynchronous
+environment, but we will use them here. Thus, all communication must have a proof that
+it maintains correctness in face of arbitrary packet reordering and dropping (via error/timeout).
+
+### Handshake
+
+Before creating that channel, you must have created the `external-staking` and
+`converter` contracts. The `external-staking` contract must be initialized with
+the valid (connection, port) from which the converter will connect, which means that
+the converter contract must be established before the `external-staking` contract.
```
maurolacy, line 61:
```suggestion
the converter contract must be instantiated before the `external-staking` contract.
```

docs/ibc/Staking.md
```diff
@@ -1,3 +1,168 @@
...
+
+We want to ensure that at any point in time, when all in-flight messages would be resolved, both the
+provider and the consumer chain have the same view of the staking state. (**TODO** Consider the
+effects of slashing, not included in this document currently). 
+
+We also want to guarantee that the provider chain always maintains sufficient staked tokens
+in the vault to cover all virtual staking actions currently outstanding on the provider chain.
+
+As [mentioned before](./ControlChannel.md#channel-ordering), we wish to use an unordered channel,
+and therefore must bring a degree of understading of [Serializability](./Serializability.md)
```
maurolacy, line 16: understading -> understanding

docs/ibc/Staking.md
```diff
@@ -1,3 +1,168 @@
...
+The consumer side wants to maintain the same counter per validator as the provider side.
+It manages a delegation count per validator - `D[V]`, with the following rules:
+
+* Uninitialized is equivalent to `D[V] = 0`
+* `S(V, X)` => `D[V] += X`, return success ack
+* `U(V, X)` => 
+  * if `D[V] < X`, return error ack
+  * else `D[V] -= X`, return success ack
+
+This is pretty straightforward counter with a lower bound of 0, along with an increment and decrement
```
maurolacy, line 30:
```suggestion
This is a pretty straightforward counter with a lower bound of 0, along with an increment and decrement
```

docs/ibc/Staking.md
```diff
@@ -1,3 +1,168 @@
...
+counters. 
+
+The only requirements to ensure this stays in sync is that the provider can successfully commit the same
+changes upon a success ack, and is able to revert then (unstake) on an error ack.
```
maurolacy, line 34: then -> them

docs/ibc/Staking.md
```diff
@@ -1,3 +1,168 @@
...
+
+## Provider Side Design
+
+Beyond this delegation list, the provider must maintain much more information.
+The delegation to a given validator must be mapped to a user on the provider chain.
+This user must also have a sufficient lien on the vault to cover the delegation.
+And the vault must have sufficient funds to cover the lien.
+
+This means we must maintain invariants over at least two contracts - the `external-staking` contract
+and the `vault`. It includes possible conflicts with non-IBC transactions, like a user to withdrawing
```
maurolacy, line 46: Remove 'to'.

docs/ibc/Staking.md
```diff
@@ -1,3 +1,168 @@
...
+* Remove a delegation in `external-staking` contract
+  * Ensure stake count (user, validator) is set and greater than desired unstake amount
+    * Ensure total stake on the validator is set and greater than desired unstake amount (should always be true if above is true)
+  * Decrease stake count for (user, validator)
+  * Decrease total stake on the validator
+  * Decrease the user's shares in the reward distribution
+  * Add an entry to the (user, validator) "pending unbonding" to withdraw said tokens after the unbonding period.
+* Send an IBC packet to inform the Consumer
+  * Guarantee we can commit all above on success
+  * Guarantee we can rollback all above on error
```
maurolacy, line 79: Looks good, but we must be really sure we're not missing some scenario / combination of operations here.

Perhaps a simpler initial approach could be, to allow for only one in-flight operation at any given time, and reject new operations until the pending one is completed, either successfully or with error.

maurolacy (commented):
Some more comments / syntax suggestions. Will review `Serializability` next.

docs/ibc/Validators.md
ethanfrey: First, that would be an illegal message (not valid with this protocol version on the sender side).
Second, the aim is to never return errors on IBC receive here for Validator Sync.
Key rotation is described in the next section

docs/ibc/Validators.md
ethanfrey: Could do that too, but seems odd to store.

The actual implementation can use different types for efficiency and scalability.
Maybe I do just write it as a Set here?

docs/ibc/Staking.md
ethanfrey: That is the locking approach.

We can't just reject other IBC packets, but any other transaction that touches the same keys.
That means, we cannot pay out rewards to that validator or make/remove another lien from that user to anyone.

That is correct, but not efficient. Please read discussion in Serializability and we take this up.

docs/ibc/Serializability.md
```diff
@@ -0,0 +1,229 @@
...
+attempt to decrease the min by 200 but leave the max unchanged (requiring this range to remain valid).
+
+This approach would only work by values not used by complex forumulas, such as balances of an AMM (we can't calculate prices off ranges),
+or paying out staking rewards (the value received by all other users depends on your stake, and we can't propogate this to all those accounts,
+as it would be prohibitively expensive to collapse them all when the transaction is committed or rolled-back).
+
+But for counters with comparisons, increments, and decrements, it could work well. Even enforcing logic like
+"total amount of collateral is greater than the max lien amount" could be enforced if collateral and lien amounts
+were value ranges. In this case, the max of (`lien.max`) would be compared against `collateral.min`.
+With some clever reasoning, we could possibly enforce such value ranges without actually storing multiple fields.
```
maurolacy: This looks like a good idea.

One concern that I have is that the value ranges will need to be cumulative. That is, a value range (or at least, part of it) of a given transaction would depend on the value range of the other un-finished transaction(s).

And if I'm not mistaken, this introduces the issue of ordering again. For each transaction, we would need a set of value ranges, reflecting each of the possible ways the sequence of pending transactions can finish.

Will think more about it.

maurolacy (commented):
More syntax suggestions and comments.

docs/ibc/Serializability.md
ethanfrey: We don't need a huge set, as it is just min, max for the range.

- Range(100, 100) + Maybe(+200) + Maybe(-50) => Range(50, 300)
- Range(50, 300) + Rollback(+200) => Range(50, 100)
- Range(50, 100) + Commit(-50) => Range(50, 50)

As I went further, I think I would use these ranges mostly as theoretical constructs to evaluate and then use locks if needed.

All queries accessing the said value would need to be range-aware, which is more complex than making them lock-aware. A lock is basically "get access or error". A range means all operations (sum, comparison, etc) would need to check multiple values.

ethanfrey: But yeah, no need for `O(N!)`

But maybe a good idea to have a set of `(caller, Maybe)` for all pending tx to validate if `Rollback` or `Commit` can be safely called.

We can keep this a tool for analysis for now and consider actual Rust implementations if needed in the protocols.

maurolacy: What about

- Range(100, 100) + Maybe(+200) + Maybe(-150) => Range(-50, 300)

?

That would fail in one order, but succeed in the other. You propose simply rejecting the offending tx in this case?

docs/ibc/Staking.md
```diff
@@ -1,3 +1,264 @@
...
+We end up with two keys that need to write locks, both of which only affect a single user:
+
+- `vault::lien(user, external-staking)`
+- `external-staking::stake(user, validator)`
+
+Note that a write-lock will prevent reading of the value from any other transaction. That means the same user offering a lien on another
+validator, which iterates over all liens to find the sum of max slashing, will be blocked until the first transaction completes.
+We cannot actually block (or wait) transactions in this way, so we must return an error for the second transaction.
+
+### Idea: Approximating Locks
```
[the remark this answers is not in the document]
maurolacy: Can't review now, but will take a look later for sure.

docs/ibc/Serializability.md
maurolacy: Whereas

 - Range(100, 100) + Maybe(-50) + Maybe(-60) => Range(-10, 100)]
 
will fail. Nice.

maurolacy: Balances and similar will also be undefined during pending transactions. They will be value ranges, basically.

docs/ibc/Staking.md
```diff
@@ -1,3 +1,252 @@
...
+which is sufficient to guarantee the invariants of the other staking values, without interfering with actions by other users.
+
+We end up with two keys that need to write locks, both of which only affect a single user:
+
+- `vault::lien(user, external-staking)`
+- `external-staking::stake(user, validator)`
+
+Note that a write-lock will prevent reading of the value from any other transaction. That means the same user offering a lien on another
+validator, which iterates over all liens to find the sum of max slashing, will be blocked until the first transaction completes.
+We cannot actually block (or wait) transactions in this way, so we must return an error for the second transaction.
```
alpe, line 142: I think that you can avoid failing the second TX by maintaining state for in flight data in both contracts.
Like credit-card companies, you could place a lock in the vault for the lien amount, that is released and the change applied on success Ack or just dropped on error Ack. The liquid balance would be reduced by the locked amounts.
Similar in external-staking, where the changes are applied only on success Ack. Or reverted on error Ack. Depending at which stage they are taken into account in the contract

alpe (commented):
I have not had a chance to read all docs but here is some feedback already

docs/ibc/Serializability.md
ethanfrey: Yes, adding the `Maybe(-150)` would fail, as that would enter a non-commutative state (where there exist different orderings that would provide different results)

docs/ibc/Staking.md
```diff
@@ -1,3 +1,252 @@
...
+It manages a delegation count per validator - `D[V]`, with the following rules:
+
+- Uninitialized is equivalent to `D[V] = 0`
+- `S(V, X)` => `D[V] += X`, return success ack
+- `U(V, X)` =>
+  - if `D[V] < X`, return error ack
+  - else `D[V] -= X`, return success ack
+
+This is a pretty straightforward counter with a lower bound of 0, along with an increment and decrement
+counters.
```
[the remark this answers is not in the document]
ethanfrey: I am very hesitant to relax invariants ever, and start handling negative numbers when an epoch hits.

The provider should not even allow the `U(a, 50)` action to be initiated on that provider side, until it is sure the `D(a, 100)` is committed. What happens if the first one fails for some reason?

With full CRDTs, we assume nothing will ever be rejected, and all operations will be applied.

With this distributed transaction model, we need to ensure consistency. Minimally, no transaction can read data from an in-flight transaction - meaning the delegation doesn't even start on the provider side until it is fully commited (on ack not on send packet)

docs/ibc/Serializability.md
ethanfrey: BTW, I started https://github.com/osmosis-labs/mesh-security/pull/52 to implement the read/write lock primitives as actual structs that can be used in `cw-storage-plus`.

If you feel inspired to start modeling value ranges, feel free. (Just let me know before you start coding in case I grab it)

docs/ibc/Staking.md
alpe: I may not have the full picture, yet but with my current understanding, the "withdraw collateral" should fail instead. The "stake" action was started before and the vault should reflect that state in the balance. With an optimistic approach, the lien should be taken on start of the "stake" process already. It can be returned on error Ack or timeout. 

I do understand the "write-lock" as "one operation only until finished". That constraint would be very hard and can be avoided IMHO. It should be possible to "withdraw collateral" and having a "stake" operation in flight as long as the balance in the vault is sufficient for both.

docs/ibc/Validators.md
```diff
@@ -0,0 +1,185 @@
...
+We define the "Validator Subprotocol" as a way to sync this information. It uses a CRDT-like
+design to maintain consistency in the face of arbitrary reordering. And retries in order
+to ensure dropped packets are eventually received.
+
+## Message Types
+
+All validator change messages are initiated from the consumer side. The provider is
+responsible for guaranteeing any packet with a success ACK is written to state.
+Given all successful acks have been committed, the consumer maintains enough
+information to sync otustanding changes and guarantee the Provider will eventually
```
alpe, line 16:
```suggestion
information to sync outstanding changes and guarantee the Provider will eventually
```

docs/ibc/Staking.md
ethanfrey: Yes, the locking approach is the easiest one to prove safe.

And from there, I look to reduce constraints, while showing that this can never open up a conflict, assuming the previous version had no conflicts.

I think you are not looking at the latest version, which has a section **Idea: Approximating Locks** discussing just what you are suggesting (and asking for feedback that this is safe as a lock)

docs/ibc/Staking.md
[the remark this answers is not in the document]
ethanfrey: @alpe Your comments above about optimisitic liens are the topic of this section.

Please read it and let me know how it can be improved.

docs/ibc/Serializability.md
ethanfrey: See #53  Man, it feels good to code again...

docs/ibc/Staking.md
```diff
@@ -1,3 +1,264 @@
...
+The consumer side wants to maintain the same counter per validator as the provider side.
+It manages a delegation count per validator - `D[V]`, with the following rules:
+
+- Uninitialized is equivalent to `D[V] = 0`
+- `S(V, X)` => `D[V] += X`, return success ack
+- `U(V, X)` =>
+  - if `D[V] < X`, return error ack
+  - else `D[V] -= X`, return success ack
+
+This is a pretty straightforward counter with a lower bound of 0, along with an increment and decrement
```
maurolacy, line 30: Perhaps reference here the upper bound (max cap) as well?

This is another issue, but I think we would need a message type for max cap to reach the provider side. That way we can know how much it can be delegated to a given remote, before sending the delegation.
That also involves converting between provider and consumer coins. So, it will have to be updated once per day or so, based on the price oracle.

Or am I mistaken here?

docs/ibc/Staking.md
```diff
@@ -1,3 +1,264 @@
...
+
+Possible keys with conflicts are:
+
+- In `vault` - `collateral(user)` and `lein(user, external-staking)`
+- In `external-staking` - `stake(user, validator)`, `total-stake(validator)`, `reward-shares(user)`, `pending-unbonding(user, validator)`
+
+### Identifying Potential Commutatibility
+
+The general design should be to write all changes only on a successful ACK, but hold any locks needed to ensure those
+write will not fail in any condition. Using the apporach of [Value Ranges](./Serializability.md#value-ranges), let us analyze
```
maurolacy, line 89: write -> writes
apporach -> approach

docs/ibc/Staking.md
```diff
@@ -1,3 +1,264 @@
...
+    }
+}
+```
+
+### Provider-Side Staking
+
+This is more complex logic over multiple contracts, so I will only give a high-level overview here rather the Rust psuedo-code:
+
+**TODO**
+See [open question above](#idea-approximating-locks) for discussion of how to possibly avoid locking on staking.
```
maurolacy, line 223:
```suggestion
See [open question above](#idea-approximating-locks) for a discussion on how to possibly avoid locking on staking.
```

maurolacy (commented):
Some syntax suggestions, and a question about max cap enforcement.

docs/ibc/Staking.md
ethanfrey: We store the total requested stake in the consumer side.

When the epoch comes, it is normalized down to max_cap if above, and then redelegated appropriately.

Eg. if max cap is 1000 and there was 500 on A and 500 B. Then I stake 1000 more on A. When the epoch tick comes, it doesn't end up 500-500 but rather 750 on A and 250 on B (staked * max cap / total staked)

ethanfrey: We let the provider do what it wants, it always has impact.
But after max_cap, there is diminishing returns per staker in terms of rewards, and they come from other cross stakers on the same channel

docs/ibc/Rewards.md
```diff
@@ -1,3 +1,64 @@
...
+is permissionless and accepts any payment in the proper denom.
+
+This is a nice solution, as it allows the converter to perform all actions in a single
+IBC packet call and not worry about orchestrating multiple packets. However, as of the
+writing of this document, there is **no callback** on failed packet receive
+(that is left for a future version of the protocol). This means that if the `distribute_rewards`
+call fails for whatever reason, the reward tokens will be effectively lost.
+
+This is a good solution for the MVP, but we should consider moving to a more robust
+solution when one exists. Hopefully this will spur deployment of such methods.
```
alpe, line 64: Another approach would be building the callback into ms-sdk with a custom message to register for the callback on any outgoing ibc package. I had outlined this in https://github.com/CosmWasm/wasmd/pull/1368. already.

packages/apis/src/ibc.rs
ethanfrey: Now the consumer can just send it in as many packets as it wants, repeating is fine, delays are fine.

docs/ibc/Staking.md
```diff
@@ -1,3 +1,264 @@
...
+what needs to be minimally enforced here.
+
+For staking, we update:
+
+- `vault::lien(user, external-staking)` => `Maybe(+X)`
+- `external-staking::stake(user, validator)` => `Maybe(+X)`
+- `external-staking::total-stake(validator)` => `Maybe(+X)`
+- `external-staking::reward-shares(validator, user)` => `Maybe(+X)`
+
+The lower three have no upper limit and thus can never fail, so those operations are always commutative with others valid operations.
```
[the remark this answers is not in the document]
ethanfrey: Well, when lots of people ape in to stake somewhere, APRs drop.
This is true all over DeFi and already (to a lesser extend) in staking rewards (in the early days of chains).
Users should be aware (or have a little disclaimer on the app)

PR #52, opened 2023-06-09
Add lock sync primitives

ethanfrey:
Based on the theoretical work in #49 

Define a `Lockable<T>` type that can be stored in an `Item` or `Map` and manages pending read or write locks in storage,
so as to hold them over multiple blocks until an IBC ack comes in.

The "normal" use now needs to be via `val.read()?` or `val.write()?` to get `&T` or `&mut T`, so a slight change to calling code, but not too bad. And there are new methods to lock or unlock it which can control whether the previous are allowed.
Also, usage in higher-level constructs (like `Map::range`) should enforce this and error if a single element has a write-lock.

### DIFF
--- a/packages/sync/src/locks.rs
+++ b/packages/sync/src/locks.rs
@@ -8,6 +8,7 @@ pub struct Lockable<T> {
 }
 
 #[cw_serde]
+#[derive(Copy)]
 pub enum LockState {
     #[serde(rename = "no")]
     Unlocked,
@@ -17,7 +18,7 @@ pub enum LockState {
     ReadLocked(u32),
 }
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq, Eq)]
 pub enum LockError {
     #[error("Value is already write locked")]
     WriteLocked,
@@ -54,6 +55,10 @@ impl<T> Lockable<T> {
             LockState::ReadLocked(_) => Err(LockError::ReadLocked),
             LockState::Unlocked => Ok(&mut self.inner),
         }
+    }
+
+    pub fn state(&self) -> LockState {
+        self.lock
     }
 
     pub fn lock_write(&mut self) -> Result<(), LockError> {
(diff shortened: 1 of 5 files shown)


packages/sync/src/locks.rs
```diff
@@ -0,0 +1,416 @@
...
+        #[error("{0}")]
+        Std(#[from] StdError),
+        #[error("{0}")]
+        Lock(#[from] LockError),
+    }
+
+    #[test]
+    fn modify_item_with_locks() {
+        let mut store = MockStorage::new();
+        const PERSON: Item<Lockable<Person>> = Item::new("person");
```
maurolacy, line 277: Interesting functionality of a persistent lock. This can prove very handy.

packages/sync/src/locks.rs
```diff
@@ -0,0 +1,416 @@
...
+                assert_eq!(p.read()?.age, 34);
+                Ok::<_, PersonError>(p)
+            })
+            .unwrap();
+    }
+
+    #[test]
+    fn map_methods_with_locks() {
+        let mut store = MockStorage::new();
+        const PEOPLE: Map<&str, Lockable<Uint128>> = Map::new("people");
```
maurolacy, line 323: Nit: Better to use other map name, like `AGES`, to avoid confusion with the previous test case.

packages/sync/src/locks.rs
```diff
@@ -0,0 +1,416 @@
...
+        // We cannot range over all
+        let err = PEOPLE
+            .range(&store, None, None, cosmwasm_std::Order::Ascending)
+            .fold(Ok(Uint128::zero()), |sum, item| {
+                Ok::<_, PersonError>(sum? + *item?.1.read()?)
+            })
+            .unwrap_err();
+        assert_eq!(err, PersonError::Lock(LockError::WriteLocked));
+
+        // We can get count (kind of edge case bug, but I don't think we can change this or it matters)
```
maurolacy, line 410: Technically, this is correct. The map itself is unlocked, and so, its state can change.

The same with removal of a locked item, I guess? Would be good to add a test for it, just to showcase it.

maurolacy (approved):
LGTM.

packages/sync/src/locks.rs
maurolacy: Unless we consider removing a (write) locked item an error, and use a wrapper around `remove` to try and enforce that.

ethanfrey: > Unless we consider removing a locked item an error, and use a wrapper around remove to try and enforce that.

Yeah, it seemed a new `LockedMap` type would be a whole lot of work, so I was trying to figure out how to do this with minimal intervention.

I think it is correct that count works, but `remove` should definitely be prohibited. Any ideas how to do so without copying over the entire Map API?

maurolacy: Rust composition over inheritance makes writing this kind of minimally modified containers a pain.

The only way I see around a `LockedMap` / `LockableMap` and similar is a `Lockable::remove` method. It feels unnatural, and enforcement will be responsibility of the user, as there's no way to avoid calling the base / original remove, but is perhaps preferable.

ethanfrey: I will add such a method as helper.
Maybe another PR to add some sort of LockableMap alternative which enforces everything

JakeHartnell (approved):
Very cool and useful package. :brain:

PR #53, opened 2023-06-09
Add ValueRange type

ethanfrey:
Requires #52 (merge that first)

This is inspired by the work in #49 about defining possible ranges of values as a lightweight partial lock for  in-flight IBC transactions

Basic functionality is there and tested. It would be good to have some more realistic tests cases to ensure this API would be useful in real conditions

### DIFF
--- a/packages/sync/src/range.rs
+++ b/packages/sync/src/range.rs
@@ -116,15 +116,16 @@ impl<T> ValueRange<T>
 where
     T: Add<Output = T> + Sub<Output = T> + Ord + Copy,
 {
-    /// This is a check for calling code if it wishes to change the (externally defined) maximum for the range.
-    /// Usage is eg modifying collateral while the range is total liens on the collateral
-    pub fn valid_max(&self, new_max: T) -> bool {
-        self.1 <= new_max
+    /// Returns true iff all values of the range are <= max.
+    /// This can be used to assert invariants
+    pub fn is_under_max(&self, max: T) -> bool {
+        self.1 <= max
     }
 
-    /// This is a check for calling code if it wishes to change the (externally defined) minimum for the range.
-    pub fn valid_min(&self, new_min: T) -> bool {
-        self.0 >= new_min
+    /// Returns true iff all values of the range are >= min.
+    /// This can be used to assert invariants
+    pub fn is_over_min(&self, min: T) -> bool {
+        self.0 >= min
     }
 
     /// This is to be called at the beginning of a transaction, to reserve the ability to commit (or rollback) an addition.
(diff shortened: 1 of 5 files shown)


maurolacy (commented):
Looks good, but it's difficult for me to see how useful this is until hitting the real use case.

What we need is a couple of values that we can keep track against a couple of min (typically, zero) and max, typically, max cap?) values.

A given set of pending unordered TXs will be valid (that means, it can be accepted without locking), if:

- All of the reduce stack ones are above the min (when / if successfully processed), in the face of all the increase stack ones failing.
- Vice versa: All of the increase stack ones are below the max (if successfully processed) in the face of all the decrease stack ones failing.

Not difficult to reason about, actually. If the `ValueRange` type helps us in that regard, that's good.

ethanfrey:
That is exactly the use case I was considering. And I think done with this API. I need to make better examples and likely adjust a bit.

A `ValueRange` keeps the min and max of all pending either succeeding or failing. We can also sum a number of them (think Sum over all liens to find total_slashable). And get the max of the max and min of the min.

It allows implementing a lock not on one element (eg user's OSMO balance), but on a portion of one element (500 OSMO of the user's balance). Want to try to write some such example as a test and give feedback on the API?

maurolacy:
I have a rough idea on how to implement this:

Based on a set of pending TXs. And (I think) we don't need locking, but just a check before accepting a new TX: It either passes (is compatible with the current set of pending TXs, in the sense of not exceeding the min and max caps). Or, it fails, and is rejected upfront.

Are you proposing using you `Lockable` type to try and define the amounts associated with the set of pending TXs? Or are you proposing to use locking if the compatibility check fails?
 
I can try implementing this on the `external-staking` contract, and we can discuss there over the PR if you want.

ethanfrey:
I think we need to agree on the IBC docs and add IBC before we start using this in contracts.

My idea is rather than have to reason though every possible transaction that touches the same contracts as an IBC transaction, the IBC transaction just locks the critical sections it will update (either a WriteLock, or a ValueRange).

Then anything that could conflict would error. In lockable, this would be a read, a write, or another lock on the same value. With ValueRange, we reserve some more of that value, and it fails if any point in the range (any combination of pending tx succeeding or failing) would violate some limit (eg below 0 or above some max cap, eg collateral)

The main point is to modify the values such that all tx will execute this lock-check code. And not having to go through every path and making sure we add checks

JakeHartnell (approved):
Looks good! Another useful package.

PR to fix clippy errors: https://github.com/osmosis-labs/mesh-security/pull/56

packages/sync/src/range.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+}
+
+impl<T> ValueRange<T>
+where
+    T: Add<Output = T> + Sub<Output = T> + Ord + Copy,
+{
+    /// This is a check for calling code if it wishes to change the (externally defined) maximum for the range.
+    /// Usage is eg modifying collateral while the range is total liens on the collateral
+    pub fn valid_max(&self, new_max: T) -> bool {
```
maurolacy, line 121: This looks good, but the API is still confusing.

I would suggest using `is`, for queries that don't modify state:
```suggestion
    pub fn is_valid_max(&self, new_max: T) -> bool {
```

packages/sync/src/range.rs
```diff
@@ -0,0 +1,399 @@
...
+        self.1 <= new_max
+    }
+
+    /// This is a check for calling code if it wishes to change the (externally defined) minimum for the range.
+    pub fn valid_min(&self, new_min: T) -> bool {
```
maurolacy, line 126:
```suggestion
    pub fn is_valid_min(&self, new_min: T) -> bool {
```

packages/sync/src/range.rs
```diff
@@ -0,0 +1,399 @@
...
+        self.0 >= new_min
+    }
+
+    /// This is to be called at the beginning of a transaction, to reserve the ability to commit (or rollback) an addition.
+    /// It doesn't enforce any maximum value. Use `prepare_add_max` for that.
+    pub fn prepare_add(&mut self, value: T) -> Result<(), RangeError> {
```
maurolacy, line 132: Is this really useful? Why not remove it, and rename `prepare_add_max` to it?

I mean, we always want to be sure we are below the max. Another option could be a couple of `set_max` / `set_min` methods. That is, storing the max as part of the type.

maurolacy (approved):
LGTM. Some comments on usability / API design, as I still find it confusing.

packages/sync/src/range.rs
ethanfrey: The purpose is places where there is no max (like token balance) as opposed to places with a max (lien against collateral).

Maybe I can add `impl Into<Option<T>>` on the end, so use would be either `val.prepare_add(150, None)` or `val.prepare_add(150, 400)`.

ethanfrey: I don't want to store the range in the type, cuz that is often stored elsewhere. If every lien stored the collateral value, I would have to iterate over all to find the max when we update the collateral. I considered using some sort of const generics, but this doesn't match usage.

I agree with combining into one to make it more explicit

packages/sync/src/range.rs
ethanfrey: Good idea

packages/sync/src/range.rs
maurolacy: I like the the one with the option more.

It's more robust, as you cannot confuse the methods, and are force to explicitly declare the max / no max.

maurolacy: For `prepare_min`, `None` would also mean "no min". Which really means "the limit imposed by the type". Which is zero for unsigned types, so, we'll both good and consistent in the API.

ethanfrey:
Okay, I made suggested improvements.
Adding another PR with examples to shake it out in real-world usage

PR #55, opened 2023-06-14
Add IBC middleware option to reward distribution

alpe:
Add ICS20 with Custom IBC Middleware section to doc


ethanfrey (approved):
Very nice addition

PR #56, opened 2023-06-14
Fix clippy errors on `value-range` branch

JakeHartnell:
Bumps toolchain to the latest rust stable version in CI. Closer to what we're likely using on our machines anyway.


ethanfrey (approved):
thanks

PR #57, opened 2023-06-15
External stacking locks

maurolacy:
Alternative to #54, but using `Lockable` for synchronisation.

Proves that `Lockable` is a useful abstraction around this.

TODO:

- [x] Send tx id in msg.
- [x] User info locking.
- [x] Commit / rollback msgs.
- [x] Adapt existing tests. 
- [x] Write tx-specific tests
  - [x] `commit_tx`.
  - [x] `rollback_tx`.
  - [x] In flight / collision.

### DIFF
--- a/contracts/provider/external-staking/src/contract.rs
+++ b/contracts/provider/external-staking/src/contract.rs
@@ -391,6 +391,7 @@ pub mod cross_staking {
             ctx: ExecCtx,
             owner: String,
             amount: Coin,
+            _tx_id: u64,
             msg: Binary,
         ) -> Result<Response, Self::Error> {
             let config = self.config.load(ctx.deps.storage)?;
--- a/contracts/provider/external-staking/src/contract.rs
+++ b/contracts/provider/external-staking/src/contract.rs
@@ -391,7 +391,7 @@ pub mod cross_staking {
             ctx: ExecCtx,
             owner: String,
             amount: Coin,
-            _tx_id: u64,
+            tx_id: u64,
             msg: Binary,
         ) -> Result<Response, Self::Error> {
             let config = self.config.load(ctx.deps.storage)?;
@@ -436,7 +436,7 @@ pub mod cross_staking {
                 .add_attribute("action", "receive_virtual_stake")
                 .add_attribute("owner", owner)
                 .add_attribute("amount", amount.amount.to_string())
-                .add_attribute("tx_id", msg.tx.to_string());
+                .add_attribute("tx_id", tx_id.to_string());
 
             Ok(resp)
         }
--- a/contracts/provider/external-staking/src/contract.rs
+++ b/contracts/provider/external-staking/src/contract.rs
@@ -435,7 +435,8 @@ pub mod cross_staking {
             let resp = Response::new()
                 .add_attribute("action", "receive_virtual_stake")
                 .add_attribute("owner", owner)
-                .add_attribute("amount", amount.amount.to_string());
+                .add_attribute("amount", amount.amount.to_string())
+                .add_attribute("tx_id", msg.tx.to_string());
 
             Ok(resp)
         }
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -284,7 +284,7 @@ fn staking() {
 #[track_caller]
 fn get_last_pending_tx_id(vault: &VaultContractProxy) -> Option<u64> {
     let txs = vault.all_pending_txs(None, None).unwrap().txs;
-    txs.last().map(|tx| tx.id)
+    txs.first().map(|tx| tx.id)
 }
 
 #[test]
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -128,9 +128,7 @@ fn staking() {
         .call(users[0])
         .unwrap();
 
-    // Hardcoded commit tx call
-    // TODO: get last tx helper
-    let mut last_tx = 1;
+    let last_tx = get_last_pending_tx_id(&vault).unwrap();
     vault
         .vault_api_proxy()
         .commit_tx(last_tx)
@@ -149,10 +147,9 @@ fn staking() {
         .call(users[0])
         .unwrap();
 
-    last_tx += 1;
     vault
         .vault_api_proxy()
-        .commit_tx(last_tx)
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
         .call(contract.contract_addr.as_str())
         .unwrap();
 
@@ -168,10 +165,9 @@ fn staking() {
         .call(users[0])
         .unwrap();
 
-    last_tx += 1;
     vault
         .vault_api_proxy()
-        .commit_tx(last_tx)
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
         .call(contract.contract_addr.as_str())
         .unwrap();
 
@@ -187,10 +183,9 @@ fn staking() {
         .call(users[1])
         .unwrap();
 
-    last_tx += 1;
     vault
         .vault_api_proxy()
-        .commit_tx(last_tx)
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
         .call(contract.contract_addr.as_str())
         .unwrap();
 
@@ -205,11 +200,9 @@ fn staking() {
         )
         .call(users[1])
         .unwrap();
-
-    last_tx += 1;
     vault
         .vault_api_proxy()
-        .commit_tx(last_tx)
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
         .call(contract.contract_addr.as_str())
         .unwrap();
 
@@ -286,6 +279,12 @@ fn staking() {
             },
         ]
     );
+}
+
+#[track_caller]
+fn get_last_pending_tx_id(vault: &VaultContractProxy) -> Option<u64> {
+    let txs = vault.all_pending_txs(None, None).unwrap().txs;
+    txs.last().map(|tx| tx.id)
 }
 
 #[test]
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -638,6 +638,12 @@ fn distribution() {
         )
         .call(users[0])
         .unwrap();
+    // Hardcoded commit_tx call (lack of IBC support yet)
+    vault
+        .vault_api_proxy()
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
+        .call(contract.contract_addr.as_str())
+        .unwrap();
 
     vault
         .stake_remote(
@@ -650,6 +656,11 @@ fn distribution() {
         )
         .call(users[0])
         .unwrap();
+    vault
+        .vault_api_proxy()
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
+        .call(contract.contract_addr.as_str())
+        .unwrap();
 
     vault
         .stake_remote(
@@ -661,6 +672,11 @@ fn distribution() {
             .unwrap(),
         )
         .call(users[1])
+        .unwrap();
+    vault
+        .vault_api_proxy()
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
+        .call(contract.contract_addr.as_str())
         .unwrap();
 
     // Start with equal distribution:
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -129,6 +129,7 @@ fn staking() {
         .unwrap();
 
     let last_tx = get_last_pending_tx_id(&vault).unwrap();
+    // Hardcoded commit_tx call (lack of IBC support yet)
     vault
         .vault_api_proxy()
         .commit_tx(last_tx)
@@ -337,6 +338,13 @@ fn unstaking() {
         )
         .call(users[0])
         .unwrap();
+    let last_tx = get_last_pending_tx_id(&vault).unwrap();
+    // Hardcoded commit_tx call (lack of IBC support yet)
+    vault
+        .vault_api_proxy()
+        .commit_tx(last_tx)
+        .call(contract.contract_addr.as_str())
+        .unwrap();
 
     vault
         .stake_remote(
@@ -349,6 +357,11 @@ fn unstaking() {
         )
         .call(users[0])
         .unwrap();
+    vault
+        .vault_api_proxy()
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
+        .call(contract.contract_addr.as_str())
+        .unwrap();
 
     vault
         .stake_remote(
@@ -360,6 +373,11 @@ fn unstaking() {
             .unwrap(),
         )
         .call(users[1])
+        .unwrap();
+    vault
+        .vault_api_proxy()
+        .commit_tx(get_last_pending_tx_id(&vault).unwrap())
+        .call(contract.contract_addr.as_str())
         .unwrap();
 
     // Properly unstake some tokens
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -1,3 +1,5 @@
+use anyhow::Result as AnyResult;
+
 use cosmwasm_std::{coin, coins, to_binary, Addr, Decimal};
 use mesh_native_staking::contract::multitest_utils::CodeId as NativeStakingCodeId;
 use mesh_native_staking::contract::InstantiateMsg as NativeStakingInstantiateMsg;
@@ -6,6 +8,8 @@ use mesh_vault::contract::multitest_utils::{CodeId as VaultCodeId, VaultContract
 use mesh_vault::contract::test_utils::VaultApi;
 use mesh_vault::msg::StakingInitInfo;
 
+use mesh_sync::Tx;
+
 use cw_multi_test::App as MtApp;
 use sylvia::multitest::App;
 
@@ -13,8 +17,6 @@ use crate::contract::cross_staking::test_utils::CrossStakingApi;
 use crate::contract::multitest_utils::{CodeId, ExternalStakingContractProxy};
 use crate::error::ContractError;
 use crate::msg::{ReceiveVirtualStake, StakeInfo};
-
-use anyhow::Result as AnyResult;
 
 const OSMO: &str = "osmo";
 const STAR: &str = "star";
@@ -285,7 +287,9 @@ fn staking() {
 #[track_caller]
 fn get_last_pending_tx_id(vault: &VaultContractProxy) -> Option<u64> {
     let txs = vault.all_pending_txs(None, None).unwrap().txs;
-    txs.first().map(|tx| tx.id)
+    txs.first().map(|tx| match tx {
+        Tx::InFlightStaking { id, .. } => *id,
+    })
 }
 
 #[test]
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -3,6 +3,7 @@ use mesh_native_staking::contract::multitest_utils::CodeId as NativeStakingCodeI
 use mesh_native_staking::contract::InstantiateMsg as NativeStakingInstantiateMsg;
 use mesh_native_staking_proxy::contract::multitest_utils::CodeId as NativeStakingProxyCodeId;
 use mesh_vault::contract::multitest_utils::{CodeId as VaultCodeId, VaultContractProxy};
+use mesh_vault::contract::test_utils::VaultApi;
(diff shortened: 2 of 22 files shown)


contracts/provider/external-staking/src/multitest.rs
```diff
@@ -132,6 +133,7 @@ fn staking() {
             contract.contract_addr.to_string(),
             coin(100, OSMO),
             to_binary(&ReceiveVirtualStake {
+                tx: 2,
```
ethanfrey, line 136: This is user provided? Seems dangerous. 

The vault should add the tx id (which it generates) in the metadata on the call to the external-staking contract.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -391,6 +391,7 @@ pub mod cross_staking {
             ctx: ExecCtx,
             owner: String,
             amount: Coin,
+            _tx_id: u64,
```
ethanfrey, line 394: This looks better here. Let's use that one (which comes from vault), rather than the one embedded in the user-provided msg.

contracts/provider/vault/src/contract.rs
```diff
@@ -274,8 +289,10 @@ impl VaultContract<'_> {
         let lienholder = ctx.deps.api.addr_validate(&lienholder)?;
 
         self.liens
-            .load(ctx.deps.storage, (&account, &lienholder))
```
ethanfrey, line 277: Nice, it doesn't look like major work to add this to the code. (And compiler points it out)

contracts/provider/vault/src/contract.rs
```diff
@@ -414,12 +433,68 @@ impl VaultContract<'_> {
...
             .may_load(ctx.deps.storage, (&ctx.info.sender, lienholder))?
-            .unwrap_or(Lien {
+            .unwrap_or(Lockable::new(Lien {
+                amount: Uint128::zero(),
+                slashable,
+            }));
+        let lien = lien.write()?;
+        lien.amount += amount;
+
+        let mut user = self
```
ethanfrey, line 443: We need a lock on user as well as (or instead of lien).

If we stake on another chain, or add/withdraw collateral, it will update user, which is a conflict.

contracts/provider/vault/src/contract.rs
```diff
@@ -414,12 +433,68 @@ impl VaultContract<'_> {
...
+            .may_load(ctx.deps.storage, (&ctx.info.sender, lienholder))?
+            .unwrap_or(Lockable::new(Lien {
                 amount: Uint128::zero(),
                 slashable,
-            });
+            }));
+        let lien = lien_lock.write()?;
         lien.amount += amount;
 
+        // TODO?: Lockable
```
ethanfrey, line 497: Yes, lockable

contracts/provider/vault/src/contract.rs
```diff
@@ -414,12 +433,68 @@ impl VaultContract<'_> {
...
+        Ok(())
+    }
+
+    /// Updates the pending txs for remote staking on any contract
+    ///
+    /// Stake (remote) is always called by the tokens owner, so the `sender` is
+    /// used as an owner address.
+    ///
+    /// Config is taken in argument as it sometimes is used outside of this function, so
+    /// we want to avoid double-fetching it
```
ethanfrey, line 469: Okay, so this is the new function to be called instead of stake.

`stake` just for local (no IBC) and `stake_tx` for remote ones (that need a callback)

contracts/provider/vault/src/contract.rs
```diff
@@ -429,11 +504,113 @@ impl VaultContract<'_> {
...
 
+        // Write lock it
+        lien_lock.lock_write()?;
         self.liens
-            .save(ctx.deps.storage, (&ctx.info.sender, lienholder), &lien)?;
+            .save(ctx.deps.storage, (&ctx.info.sender, lienholder), &lien_lock)?;
 
         self.users.save(ctx.deps.storage, &ctx.info.sender, &user)?;
 
+        // Create new tx
```
ethanfrey, line 514: Good to save them here.

contracts/provider/vault/src/contract.rs
```diff
@@ -429,11 +504,113 @@ impl VaultContract<'_> {
 
         ensure!(user.verify_collateral(), ContractError::InsufficentBalance);
 
+        // Write lock it
+        lien_lock.lock_write()?;
         self.liens
-            .save(ctx.deps.storage, (&ctx.info.sender, lienholder), &lien)?;
+            .save(ctx.deps.storage, (&ctx.info.sender, lienholder), &lien_lock)?;
```
ethanfrey, line 510: Note: We write lock lien and user (and save that lock), so they will not change during the IBC transaction.

There are two approaches, you seem to take the first, but let's make this explicit decision (and it should be documented at least in the rustdoc).

1. Write lock lien and user and update the values (checking consistency) in the beginning. On success, we remove pending tx and the lock. On failure, we subtract the staked value from the items then unlock. (Unlock requires a read and write).
2. Write lock lien and user and check consistency (but don't update) in the beginning. On success, we apply the pending tx, remove it and unlock the state. On rollback, we remove the pending tx and unlock, but don't touch state changes.

For here they are both fine, and happy with you doing 1. For things like unstaking, which adds a new item to a queue of pending unbonding, I would write lock the proper items and only add the pending unbonding on the success ack, as it gets a bit difficult to track which claim to delete. (Easier to write than undo).

contracts/provider/vault/src/txs.rs
```diff
@@ -0,0 +1,67 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, Decimal, Order, StdResult, Storage, Uint128};
+use cw_storage_plus::{Index, IndexList, IndexedMap, MultiIndex};
+
+#[cw_serde]
+pub enum TxType {
+    Stake,
+    Unstake,
```
ethanfrey, line 8: Unstake is never pending here.
It is triggered on the external-staking contract.
When it goes back to the vault, that is a local call.

We just have one variant, so you may just want to rename the type to InFlightStakingTx or something a little less verbos but still descriptive.

contracts/provider/vault/src/txs.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub amount: Uint128,
+    /// Slashable portion of lien
+    pub slashable: Decimal,
+    /// Associated user
+    pub user: Addr,
+    /// Remote staking contract
+    pub lienholder: Addr,
+}
+
+pub struct TxIndexes<'a> {
```
ethanfrey, line 29: why do we need these indexes? just for queries?

contracts/provider/vault/src/contract.rs
```diff
@@ -429,11 +504,113 @@ impl VaultContract<'_> {
...
+    /// Commits a pending tx
+    // TODO: Add callback handler
+    #[allow(unused)]
+    fn commit_tx(&self, ctx: &mut ExecCtx, tx_id: u64) -> Result<(), ContractError> {
+        // Load tx
+        let tx = self.pending.txs.load(ctx.deps.storage, tx_id)?;
+        // TODO: Properly handle tx type
+        assert!(tx.ty == TxType::Stake);
+        // Verify tx comes from the right contract
+        ensure!(
```
ethanfrey, line 539: good check

contracts/provider/vault/src/contract.rs
```diff
@@ -429,11 +504,113 @@ impl VaultContract<'_> {
...
+            tx.lienholder == ctx.info.sender,
+            ContractError::WrongContractTx(tx_id, ctx.info.sender.clone())
+        );
+
+        // Load lien
```
ethanfrey, line 544: Yup, looks proper

ethanfrey: (You will also want to unlock the user, as that should be locked in the first part)

contracts/provider/vault/src/contract.rs
```diff
@@ -429,11 +504,113 @@ impl VaultContract<'_> {
...
+        )?;
+
+        // Remove tx
+        self.pending.txs.remove(ctx.deps.storage, tx_id)?;
+
+        Ok(())
+    }
+
+    /// Rollbacks a pending tx
+    // TODO: Add callback handler
```
ethanfrey, line 564: What do you mean by callback handler? An ExecuteMsg variant?

contracts/provider/vault/src/contract.rs
```diff
@@ -429,11 +504,113 @@ impl VaultContract<'_> {
...
+        lien.amount -= tx.amount;
+        // Save it unlocked
+        self.liens.save(
+            ctx.deps.storage,
+            (&ctx.info.sender, &tx.lienholder),
+            &lien_lock,
+        )?;
+
+        // Rollback user's max_lien
+        let mut user = self.users.load(ctx.deps.storage, &tx.user)?;
```
ethanfrey, line 593: Yeah, this is kind of why I wouldn't apply it in the beginning, just lock and only apply on the success case.

Calculating maximum is harder when reducing (are there others higher?) vs raising (compare my new value with the max). This is an argument in favor of approach 2 above

contracts/provider/vault/src/contract.rs
```diff
@@ -447,36 +624,48 @@ impl VaultContract<'_> {
...
         // is already written to storage
         user.max_lien = self
             .liens
             .prefix(&owner)
             .range(ctx.deps.storage, None, None, Order::Ascending)
-            .try_fold(Uint128::zero(), |max_lien, lien| {
-                lien.map(|(_, lien)| max_lien.max(lien.amount))
+            .try_fold(Uint128::zero(), |max_lien, item| {
+                let (_, lien_lock) = item?;
+                // FIXME: Fails as write locked
```
ethanfrey, line 651: Huh? This only fails if another transaction by that user is in-flight, right? (Only case of the write lock). That is actually a good case to fail in (and force a client-side retry)

packages/mocks/src/cross_staking.rs
```diff
@@ -120,7 +121,7 @@ impl CrossStakingApi for MockCrossStakingContract<'_> {
         );
 
         // ignore args
-        let _ = (owner, msg, amount);
+        let _ = (owner, msg, amount, tx_id);
```
ethanfrey, line 124: Proper mock should have a way to trigger callbacks to test.

Add some ExecuteMsg variant (unpermissioned for mock testing) that calls either commit_tx or rollback_tx on the vault to test both cases (and conflicting tx in the inflight time)

ethanfrey (commented):
Looks like a good start. Add some detailed comments, but it is really minor stuff.

Updating the external-staking (and mock-external-staking) contracts is an important step for tests. The real one should just send back a message in the same tx to commit_tx. When there is IBC, we can change it, but at least you upgrade the vault <-> external staking communication to be "IBC safe"

contracts/provider/vault/src/contract.rs
maurolacy: Work in progress. Will look better soon.

contracts/provider/vault/src/contract.rs
maurolacy: Yup. We'll need to extend the messages with commit and rollback tx. Will do as part of this PR, so that I can make tests work.

contracts/provider/vault/src/contract.rs
maurolacy: Yeah, that looks like an outdsted comment. Will address tomorrow.

Thanks for the early review.

contracts/provider/vault/src/contract.rs
maurolacy: Thinking on unifying those, as the logic is the same, plus the write lock and tx for remote staking.

contracts/provider/vault/src/txs.rs
maurolacy: Yes, this was carried over from the previous impl.  I think it makes sense to keep it, for future queries.

contracts/provider/vault/src/txs.rs
maurolacy: Oh, you're right. Was hastily generalising. Will remove.

contracts/provider/vault/src/contract.rs
maurolacy: Doing 1. has the advantage that the local vs remote code looks pretty similar.

Given that rollback would be rarely called, I would leave it as it is now.

contracts/provider/external-staking/src/multitest.rs
maurolacy: Fixed.

contracts/provider/vault/src/contract.rs
```diff
@@ -47,25 +49,36 @@ pub struct VaultContract<'a> {
...
             local_staking: Item::new("local_staking"),
             liens: Map::new("liens"),
             users: Map::new("users"),
+            pending: Txs::new("pending_txs", "users"),
+            tx_count: Item::new("tx_count"),
         }
     }
 
+    pub fn next_tx_id(&self, store: &mut dyn Storage) -> StdResult<u64> {
+        let id: u64 = self.tx_count.may_load(store)?.unwrap_or_default() + 1;
```
ethanfrey, line 77: Note: this may be the best use of update. You can use it as a return result directly.
(But it may require initializing the count to 0 in instantiate)

contracts/provider/vault/src/contract.rs
```diff
@@ -98,12 +111,14 @@ impl VaultContract<'_> {
         let denom = self.config.load(ctx.deps.storage)?.denom;
         let amount = must_pay(&ctx.info, &denom)?;
 
-        let mut user = self
```
ethanfrey, line 101: These non-ibc ones look like straightforward changes

contracts/provider/vault/src/contract.rs
```diff
@@ -332,32 +359,76 @@ impl VaultContract<'_> {
...
-                        .as_ref()
-                        .map(|(_, account)| account.collateral.is_zero())
-                        .unwrap_or(false))
+                account
+                    .as_ref()
+                    .map(|(_, account_lock)| {
+                        account_lock
+                            .read()
+                            .map(|account| !with_collateral || !account.collateral.is_zero()) // Skip zero collateral
+                            .unwrap_or(false) // Skip write-locked accounts
```
ethanfrey, line 372: I question skipping and not just returning them as locked.

Then again, let's not spend time here... this is only used for clients - let's get logic working and take feedback from clients before modifying. Just wanted to note

contracts/provider/vault/src/contract.rs
```diff
@@ -332,32 +359,76 @@ impl VaultContract<'_> {
...
+        start_after: Option<u64>,
+        limit: Option<u32>,
+    ) -> Result<AllTxsResponse, ContractError> {
+        let limit = clamp_page_limit(limit);
+        let bound = start_after.and_then(Bounder::exclusive_bound);
+
+        let txs = self
+            .pending
+            .txs
+            .range(ctx.deps.storage, None, bound, Order::Descending)
```
ethanfrey, line 419: Ah, these are ordered descending. That is not the most obvious sort order, so maybe good to include in the method name somehow

contracts/provider/vault/src/contract.rs
```diff
@@ -404,36 +475,145 @@ impl VaultContract<'_> {
         lienholder: &Addr,
         slashable: Decimal,
         amount: Coin,
-    ) -> Result<(), ContractError> {
+        remote: bool,
```
ethanfrey, line 478: A bit odd to see true/false when calling. Maybe worth some `enum Staker { Remote, Local }` type for clarity. Or maybe I am overthinking.

contracts/provider/vault/src/contract.rs
```diff
@@ -404,36 +475,145 @@ impl VaultContract<'_> {
...
+            tx.lienholder == ctx.info.sender,
+            ContractError::WrongContractTx(tx_id, ctx.info.sender.clone())
+        );
+
+        // Load lien
+        let mut lien_lock = self
+            .liens
+            .load(ctx.deps.storage, (&tx.user, &tx.lienholder))?;
+        // Unlock it
+        lien_lock.unlock_write()?;
```
ethanfrey, line 554: Looks correct

contracts/provider/vault/src/contract.rs
```diff
@@ -404,36 +475,145 @@ impl VaultContract<'_> {
...
+
+        // Load user
+        let mut user_lock = self.users.load(ctx.deps.storage, &tx.user)?;
+        // Rollback user's max_lien (need to unlock it first)
+        user_lock.unlock_write()?;
+        let mut user = user_lock.write()?;
+
+        // Max lien has to be recalculated from scratch; the just rolled back lien
+        // is already written to storage
+        user.max_lien = self
```
ethanfrey, line 601: Nit: this is used elsewhere, right, maybe worth pulling into one function if it is the same.

This is complex enough logic it is good to be tested more and not have two implementations floating, that might get out of sync

contracts/provider/vault/src/contract.rs
```diff
@@ -447,36 +627,48 @@ impl VaultContract<'_> {
         let amount = amount.amount;
 
         let owner = Addr::unchecked(owner);
-        let mut lien = self
+        let mut lien_lock = self
             .liens
             .may_load(ctx.deps.storage, (&owner, &ctx.info.sender))?
             .ok_or(ContractError::UnknownLienholder)?;
+        let lien = lien_lock.read()?;
```
ethanfrey, line 634: Why do read then later write, rather than just doing a `write` lock at the first time?

contracts/provider/vault/src/contract.rs
```diff
@@ -447,36 +627,48 @@ impl VaultContract<'_> {
...
         // is already written to storage
         user.max_lien = self
             .liens
             .prefix(&owner)
             .range(ctx.deps.storage, None, None, Order::Ascending)
-            .try_fold(Uint128::zero(), |max_lien, lien| {
-                lien.map(|(_, lien)| max_lien.max(lien.amount))
+            .try_fold(Uint128::zero(), |max_lien, item| {
+                let (_, lien_lock) = item?;
+                let lien = lien_lock.read().map_err(Into::<ContractError>::into);
```
ethanfrey, line 655: Yup, duplicated above. Worth making a function imo

contracts/provider/vault/src/contract.rs
```diff
@@ -527,4 +719,27 @@ impl VaultApi for VaultContract<'_> {
 
         Ok(resp)
     }
+
+    #[msg(exec)]
```
ethanfrey, line 723: Safe to expose as they assert the caller. Nice design

ethanfrey (approved):
General design looks good.
Added a number of smaller style comments for cleanup.
It would be good to see integration into external-staking, but even more, we need to start IBC and figure out how to test well.

contracts/provider/vault/src/txs.rs
```diff
@@ -0,0 +1,66 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, Decimal, Order, StdResult, Storage, Uint128};
+use cw_storage_plus::{Index, IndexList, IndexedMap, MultiIndex};
+
+#[cw_serde]
+pub enum TxType {
+    InFlightStaking,
+    // TODO
+    // Slash,
```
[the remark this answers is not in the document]
maurolacy: Good point. Will refactor.

contracts/provider/vault/src/contract.rs
maurolacy: Copied this from somewhere else. Will refactor if possible.

contracts/provider/vault/src/contract.rs
maurolacy: Just a silly "optimisation" in case the test below fails. Will change.

contracts/provider/vault/src/contract.rs
maurolacy: Will add a FIXME, or fit it directly.

ethanfrey: THere is no clear good solution.
Just add a `// FIXME` and we figure it out when everything is finished

contracts/provider/vault/src/contract.rs
```diff
@@ -404,36 +475,145 @@ impl VaultContract<'_> {
...
+        }
+        self.liens
+            .save(ctx.deps.storage, (&ctx.info.sender, lienholder), &lien_lock)?;
+
+        self.users
+            .save(ctx.deps.storage, &ctx.info.sender, &user_lock)?;
+
+        if remote {
+            // Create new tx
+            let tx_id = self.next_tx_id(ctx.deps.storage)?;
```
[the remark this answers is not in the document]
maurolacy: Because the saves have to be processed in both branches. I think it's clearer this way, yes.

contracts/provider/vault/src/contract.rs
maurolacy: I think it's OK. A little old style sometimes is not a bad thing. :-)

contracts/provider/vault/src/contract.rs
maurolacy: Refactored / renamed for clarity.

PR #58, opened 2023-06-17
Add extra tests / examples of lockable ranges

maurolacy:
Some more tests / examples of using our new `Lockable` primitive in range iterators.

Found useful writing these to reason about and learn how to use them.

- Range iterator returning a collection.
- Locked values re-mapping.
- Locked values skipping.

### DIFF
--- a/packages/sync/src/locks.rs
+++ b/packages/sync/src/locks.rs
@@ -264,11 +264,22 @@ mod tests_plus {
     }
 
     #[derive(Error, Debug, PartialEq)]
-    pub enum PersonError {
+    pub enum TestsError {
         #[error("{0}")]
         Std(#[from] StdError),
         #[error("{0}")]
         Lock(#[from] LockError),
+    }
+
+    #[cw_serde]
+    #[derive(Default)]
+    pub struct UserInfo {
+        // User collateral
+        pub collateral: Uint128,
+        // Highest user lien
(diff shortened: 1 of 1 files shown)


packages/sync/src/locks.rs
```diff
@@ -398,15 +409,140 @@ mod tests_plus {
...
+                        .read()
+                        .map(|user| Ok(user.collateral))
+                        .unwrap_or(Ok(Uint128::zero())) // Re-map locked collateral
+                })? // Surface errors
+            })
+            .collect::<Result<_, TestsError>>()
+            .unwrap();
+        assert_eq!(collaterals.len(), 3);
+
+        // Or we can skip (perhaps not a good idea either) the write-locked values
```
ethanfrey, line 527: Reflecting more on this, I think the best would be to return:

```rust
enum MaybeAccount {
  Account(Account),
  Locked{ user: String },
}
```

Or something Similar. (Not sure how you use the keys).
I can make an example as follow up PR (unless this makes sense to you and you want to try)

ethanfrey (commented):
One more idea after merging...

packages/sync/src/locks.rs
maurolacy: 👍🏼 Will implement this for reference / example.

PR #59, opened 2023-06-19
Real-world examples of value range

ethanfrey:
Builds on #53 

- Add user / lien example
- Fix serde-json-wasm compatibility issue
- Show how to range, reduce, enforce maximums over many

### DIFF
--- a/packages/sync/src/range.rs
+++ b/packages/sync/src/range.rs
@@ -494,32 +494,30 @@ mod examples {
 
     const USERS: Map<&str, UserInfo> = Map::new("users");
 
-    // sum them up for each user - collateral info must be provided externally
-    fn sum_liens(
+    // Recalculate user aggregate info from the liens (and the original user collateral)
+    // Question: return (Vec<_>, Vec<_>) here instead and don't pass in collateral?
+    fn collect_liens_for_user(
         storage: &dyn Storage,
         user: &str,
         collateral: impl Into<Uint128>,
         // if set, we don't include this lien in the sum (to use when reducing one)
         skip: Option<&str>,
     ) -> Result<UserInfo, TestsError> {
-        let (max_lien, total_slashable) = LIENS
+        // Create (amount, slashable) extract from all liens (possibly skipping one)
+        let all_liens = LIENS
             .prefix(user)
             .range(storage, None, None, Order::Ascending)
             .filter(|r| match (skip, r) {
                 (Some(x), Ok((k, _))) => x != k,
                 _ => true,
             })
-            .map(|r| {
-                let (_, lien) = r?;
-                Ok((lien.amount, lien.amount * lien.slashable))
-            })
-            .reduce(|acc: Result<_, TestsError>, x| {
-                let (max_lien, total_slashable) = acc?;
-                let (amount, slashable) = x?;
-                Ok((max_range(max_lien, amount), total_slashable + slashable))
-            })
-            .transpose()?
-            .unwrap_or_default();
+            .map(|r| r.map(|(_, lien)| (lien.amount, lien.amount * lien.slashable)));
+
+        // collect into two vectors, which we can reduce / sum independently
+        let (liens, slashable): (Vec<_>, Vec<_>) =
+            itertools::process_results(all_liens, |iter| iter.unzip())?;
+        let max_lien = liens.into_iter().reduce(max_range).unwrap_or_default();
+        let total_slashable = slashable.into_iter().sum();
         Ok(UserInfo {
             collateral: collateral.into(),
             max_lien,
@@ -595,15 +593,15 @@ mod examples {
             )
             .unwrap();
 
-        let mut alice_user = sum_liens(&store, alice, alice_collateral, None).unwrap();
+        let mut alice_user = collect_liens_for_user(&store, alice, alice_collateral, None).unwrap();
         assert!(alice_user.is_valid());
         USERS.save(&mut store, alice, &alice_user).unwrap();
 
-        let bob_user = sum_liens(&store, bob, bob_collateral, None).unwrap();
+        let bob_user = collect_liens_for_user(&store, bob, bob_collateral, None).unwrap();
         assert!(bob_user.is_valid());
         USERS.save(&mut store, bob, &bob_user).unwrap();
 
-        let mut carl_user = sum_liens(&store, carl, carl_collateral, None).unwrap();
+        let mut carl_user = collect_liens_for_user(&store, carl, carl_collateral, None).unwrap();
         assert!(carl_user.is_valid());
         USERS.save(&mut store, carl, &carl_user).unwrap();
 
@@ -650,7 +648,8 @@ mod examples {
         LIENS.save(&mut store, (alice, stake1), &lien).unwrap();
         USERS.save(&mut store, alice, &alice_user).unwrap();
         // verify this matches the full calculation
-        let alice_user2 = sum_liens(&store, alice, alice_user.collateral, None).unwrap();
+        let alice_user2 =
+            collect_liens_for_user(&store, alice, alice_user.collateral, None).unwrap();
         assert_eq!(alice_user, alice_user2);
 
         // but 500 more is too much
@@ -666,7 +665,8 @@ mod examples {
             .prepare_sub(Uint128::new(2000), Uint128::zero())
             .unwrap();
         // this requires a bit more tricky version to get max_lien
-        let mut bob_user = sum_liens(&store, bob, bob_user.collateral, Some(stake1)).unwrap();
+        let mut bob_user =
+            collect_liens_for_user(&store, bob, bob_user.collateral, Some(stake1)).unwrap();
         // and add this lien fully
         bob_user.total_slashable = bob_user.total_slashable + (lien.amount * lien.slashable);
         bob_user.max_lien = max_range(bob_user.max_lien, lien.amount);
@@ -676,7 +676,7 @@ mod examples {
         LIENS.save(&mut store, (bob, stake1), &lien).unwrap();
         USERS.save(&mut store, bob, &bob_user).unwrap();
         // verify this matches the full calculation
-        let bob_user2 = sum_liens(&store, bob, bob_user.collateral, None).unwrap();
+        let bob_user2 = collect_liens_for_user(&store, bob, bob_user.collateral, None).unwrap();
         assert_eq!(bob_user, bob_user2);
     }
 }
--- a/packages/sync/src/range.rs
+++ b/packages/sync/src/range.rs
@@ -488,6 +488,10 @@ mod examples {
         let stake1 = "Stake1";
         let stake2 = "Stake2";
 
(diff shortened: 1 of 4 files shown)


packages/sync/src/range.rs
```diff
@@ -9,67 +9,108 @@ use cosmwasm_schema::cw_serde;
 /// This is designed to work with two numeric primitives that can be added, subtracted, and compared.
 #[cw_serde]
 #[derive(Default, Copy)]
-pub struct ValueRange<T>(T, T);
+// Note: this was (T, T) but (Uint128, Uint128) hit `serialize_tuple_struct` which is not supported by serde-json-wasm
```
maurolacy, line 12: A struct is in general better for clarity / documentation. Let's just remove this comment.

packages/sync/src/range.rs
```diff
@@ -9,67 +9,108 @@ use cosmwasm_schema::cw_serde;
...
     Underflow,
     #[error("Overflow maximum value")]
     Overflow,
+    #[error("Range is not one value")]
+    NotOneValue,
+}
+
+impl<T> ValueRange<T> {
+    /// Constructor as close to the old tuple
+    /// ValueRange(5, 10) => ValueRange::at(5, 10)
```
maurolacy, line 32: Same here.

packages/sync/src/range.rs
```diff
@@ -9,67 +9,108 @@ use cosmwasm_schema::cw_serde;
...
+    #[inline]
+    pub fn at(min: T, max: T) -> Self {
```
maurolacy, line 34: Nit: This can probably be named `new`. And `new` below can be `new_val`.

maurolacy (commented):
Reviewed the API. Will review the examples next.

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+        // Highest user lien
+        pub max_lien: ValueRange<Uint128>,
+        // Total slashable amount for user
+        pub total_slashable: ValueRange<Uint128>,
+    }
+
+    impl UserInfo {
+        fn is_valid(&self) -> bool {
+            self.max_lien.is_under_max(self.collateral)
+                && self.total_slashable.is_under_max(self.collateral)
```
maurolacy, line 484: 👍🏼 clear and concise.

Still, it can be a good to differentiate which one is failing, i.e. a couple of 
`is_valid_max_lien` and `is_valid_slashable` methods for clarity.

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+        storage: &dyn Storage,
+        user: &str,
+        collateral: impl Into<Uint128>,
+        // if set, we don't include this lien in the sum (to use when reducing one)
+        skip: Option<&str>,
+    ) -> Result<UserInfo, TestsError> {
+        let (max_lien, total_slashable) = LIENS
+            .prefix(user)
+            .range(storage, None, None, Order::Ascending)
+            .filter(|r| !matches!((skip, r), (Some(x), Ok((k, _))) if x == k))
```
maurolacy, line 511: Ugh. This is ugly.

Using `item` instead of `r` doesn't help either. There should be a way to write this in a clearer / more expressive way.

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+            .map(|r| {
+                let (_, lien) = r?;
+                Ok((lien.amount, lien.amount * lien.slashable))
+            })
+            .reduce(|acc: Result<_, TestsError>, x| {
```
maurolacy, line 516:
```suggestion
            .reduce(|acc: Result<(ValueRange<_>, ValueRange<_>), TestsError>, x| {
```
for clarity?

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+                let (max_lien, total_slashable) = acc?;
+                let (amount, slashable) = x?;
+                Ok((max_range(max_lien, amount), total_slashable + slashable))
```
maurolacy, line 519: I understand it, but this is neither user-friendly nor obvious.

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+        /// Slashable part - restricted to [0; 1] range
+        pub slashable: Decimal,
+    }
+
+    const LIENS: Map<(&str, &str), Lien> = Map::new("liens");
+
+    const USERS: Map<&str, UserInfo> = Map::new("users");
+
+    // sum them up for each user - collateral info must be provided externally
+    fn sum_liens(
```
maurolacy, line 501: This is not a sum, but more of an overlap, right? Perhaps something like `max_lien`?

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+                let (_, lien) = r?;
+                Ok((lien.amount, lien.amount * lien.slashable))
+            })
+            .reduce(|acc: Result<_, TestsError>, x| {
+                let (max_lien, total_slashable) = acc?;
+                let (amount, slashable) = x?;
+                Ok((max_range(max_lien, amount), total_slashable + slashable))
+            })
+            .transpose()?
+            .unwrap_or_default();
```
maurolacy, line 522: Asserting that `max_lien` (and `total_slashable`?) are below `collateral` here could be a good idea. Perhaps returning a differentiated error.

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+        let bob_collateral = Uint128::new(4000);
+        let carl_collateral = Uint128::new(7000);
+
+        // no inflight transactions for Alice
+        LIENS
+            .save(
+                &mut store,
+                (alice, stake1),
+                &Lien {
+                    amount: ValueRange::at(Uint128::new(5000), Uint128::new(5000)),
```
maurolacy, line 550: Nit: You can use the other constructor here and elsewhere.

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+                },
+            )
+            .unwrap();
+        // add a few liens with inflight transactions
+        LIENS
+            .save(
+                &mut store,
+                (carl, stake1),
+                &Lien {
+                    amount: ValueRange::at(Uint128::new(1000), Uint128::new(2000)),
```
maurolacy, line 582: Hmm, why would the min be different from zero here?

packages/sync/src/range.rs
```diff
@@ -387,7 +447,236 @@ mod tests {
...
+                    slashable: Decimal::percent(50),
+                },
+            )
+            .unwrap();
+        LIENS
+            .save(
+                &mut store,
+                (carl, stake2),
+                &Lien {
+                    amount: ValueRange::at(Uint128::new(5000), Uint128::new(6000)),
```
maurolacy, line 592: Same here.

maurolacy (approved):
Looks good. Just some comments on naming, and improving the general structure of tests / checks.

I would put the validity check inside the reducing function, error there with different errors, and assert those error codes in the tests for clarity.

maurolacy:
Adding an example for unstacking, where checks are below the min, would be good as well.

Will revisit during the day, but at first sight, the `ValueRange` API / stuff is not trivial.

packages/sync/src/range.rs
ethanfrey: There is, but clippy told me to do this.

Simplest and longest:

```rust
if let Some(x) = skip {
  if let Ok((k, _)) = r {
    if x == k {
      return false;
    }
  }
}
return true;
```

My version with matches:

```rust
matches (skip, r) {
  (Some(x), Ok((k, _)) => x != k,
  _ => true
}
```

And then clippy told me to turn mine into that...

packages/sync/src/range.rs
ethanfrey: The largest problem is reducing over a bunch of Results.
The second issue is I am doing two independent calculations.

```rust
.reduce(|max_lien, lien| {
  max_range(max_lien, lien.amount)
}
```

This would be one calculation with no result and look more clear, right?
Happy for a way to improve it without iterating twice....

packages/sync/src/range.rs
ethanfrey: Yes, again this is test code so I don't know if naming is that critical, but if you want to port it to contracts it is.

We do two aggregation functions. What would you call them? What do you call this in your real code?

packages/sync/src/range.rs
ethanfrey: Imagine there was a successful 1000 stake.
A day later, they try another 1000 that is in-flight.

I am just trying to get numbers that are a bit more complex. And this is a realistic value
(Not so realistic to have so many in-flight transactions, but still...)

packages/sync/src/range.rs
ethanfrey: okay

packages/sync/src/range.rs
ethanfrey: good idea

packages/sync/src/range.rs
ethanfrey: I rather one validate method to check.
This could return an error enum (with multiple reasons) instead of a bool for more clarity.
But I would leave that for production code

packages/sync/src/range.rs
ethanfrey: Okay, this worked. Clippy accepts it and it seems clear and concise:

```rust
matches (skip, r) {
  (Some(x), Ok((k, _)) => x != k,
  _ => true
}
```

packages/sync/src/range.rs
ethanfrey: I wouldn't have used any type, but the compiler could not infer the error type, so I had to add it

ethanfrey: My IDE shows me the types (at least when deconstructing acc)

packages/sync/src/range.rs
maurolacy: So, these ranges are absolute, not relative to the current tx. That makes sense.

packages/sync/src/range.rs
maurolacy: Still stuck with IntelliJ, which currently cannot follow this. Switching back and forth to Visual Studio Code.

packages/sync/src/range.rs
maurolacy: `max_overlapping_lien`, `max_slashable_amount`, something like that. Using helper functions could be a way to break from the nested scopes and provide some clarity.

packages/sync/src/range.rs
ethanfrey: Check out the last commit. I made this much simpler

ethanfrey: @maurolacy https://github.com/osmosis-labs/mesh-security/pull/59/commits/28bb0ee920378a97388698b697f8315bab26f427

This should be nicer, right?

PR #60, opened 2023-06-20
External staking locks 2

maurolacy:
#57 follow-up. Implementation of staking locks for the external-staking contract.

TODO:

- [x] Staking commit / rollback.
- [x] Unstake commit / rollback.
- [x] Adjust tests.
- [x] Send IBC packets
- [x] Route IBC ack/timeout to commit / rollback methods.
- [ ] Write tx-specific tests.
- [ ] Use a proper testing mechanism for commit / rollback (IBC messages).


contracts/provider/external-staking/src/msg.rs
```diff
@@ -94,3 +94,11 @@ pub struct UsersResponse {
 pub struct PendingRewards {
     pub amount: Coin,
 }
+
+pub type TxResponse = mesh_sync::Tx;
+pub type AllTxsResponseItem = TxResponse;
```
ethanfrey, line 99: I understand the above type definition, but why this one?

contracts/provider/external-staking/src/multitest.rs
```diff
@@ -132,14 +132,20 @@ fn staking() {
...
     vault
         .vault_api_proxy()
         .commit_tx(last_tx)
         .call(contract.contract_addr.as_str())
         .unwrap();
+    // TODO: Hardcoded external-staking's commit_stake call (lack of IBC support yet).
+    // This should be through `IbcPacketAckMsg`
+    let last_external_staking_tx = get_last_external_staking_pending_tx_id(&contract).unwrap();
+    contract
+        .commit_stake(last_external_staking_tx)
```
ethanfrey, line 146: This should be turned into a Sudo message, so we can use it for testing, but it is not there to be misused by any real user. I see keeping it so we can test most logic without IBC.

(Maybe two sudo messages, one simulates receiving an IBC ack success, other IBC ack error or timeout - and then they call into the same logic as the ibc code, except some basic parsing / formating logic

contracts/provider/vault/src/multitest.rs
```diff
@@ -451,6 +451,8 @@ fn get_last_pending_tx_id(vault: &VaultContractProxy) -> Option<u64> {
     let txs = vault.all_pending_txs_desc(None, None).unwrap().txs;
     txs.first().map(|tx| match tx {
         Tx::InFlightStaking { id, .. } => *id,
+        Tx::InFlightRemoteStaking { id, .. } => *id,
+        Tx::InFlightRemoteUnstaking { id, .. } => *id,
```
ethanfrey, line 455: Nit. I think it is easier to not include the other enum now.
And only add it when you want to implement it, so the compiler shows you all the places that need the updates

(easier than searching for unimplemented!())

packages/sync/src/txs.rs
```diff
@@ -16,8 +16,29 @@ pub enum Tx {
...
+        /// Transaction id
+        id: u64,
+        /// Associated amount
+        amount: Uint128,
+        /// Associated owner
+        user: Addr,
+        /// Remote validator
+        validator: String,
+    },
+    // IBC flight
```
ethanfrey, line 30: Better rustdoc here and above?

packages/sync/src/txs.rs
```diff
@@ -33,6 +54,30 @@ impl std::fmt::Display for Tx {
...
             }
+            Tx::InFlightRemoteStaking {
+                id,
+                amount,
+                user,
+                validator,
+            } => {
+                write!(
+                    f,
+                    "InFlightRemoteStaking {{ id: {}, amount: {}, user: {}, validator: {} }}",
```
ethanfrey, line 65: This is basically the Debug format, right?
I would assume Display to do something more custom. Or just delegate to Debug.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +98,122 @@ impl ExternalStakingContract<'_> {
         Ok(Response::new())
     }
 
+    /// Commits a pending stake.
+    /// Must be called by the IBC callback handler on successful remote staking.
+    #[msg(exec)] // TODO: Only enable for tests / use another method (IBC message)
```
ethanfrey, line 103: I propose:

1. pull this logic into it's own (undecorated) method
2. add a SudoMsg variant to call this logic (taking tx_id as arg)
3. IBC ack logic will check error or success and on success call the same method

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +98,122 @@ impl ExternalStakingContract<'_> {
...
+        // Remove tx
+        self.pending_txs.remove(ctx.deps.storage, tx_id);
+
+        // TODO: Call commit hook on vault
+
+        Ok(Response::new())
+    }
+
+    /// Rollbacks a pending stake.
+    /// Must be called by the IBC callback handler on failed remote staking.
```
ethanfrey, line 164: Same as above comment on commit_stake: using Sudo (for testing) and IBC ack logic wrap one method that does this logic

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +98,122 @@ impl ExternalStakingContract<'_> {
...
+            _ => unreachable!(),
+        };
+
+        // Load stake
+        let mut stake_lock = self
+            .stakes
+            .load(ctx.deps.storage, (&tx_user, &tx_validator))?;
+
+        // Load distribution
+        let mut distribution_lock = self.distribution.load(ctx.deps.storage, &tx_validator)?;
```
ethanfrey, line 131: Do not lock distribution please.

I described this in [the reward doc](https://github.com/osmosis-labs/mesh-security/blob/main/docs/ibc/Rewards.md#reward-distribution)

> The further question arises about how to count people with in-flight staking (or unstaking) transactions at the time of the epoch reward packet arriving. This is a bit more tricky, as we don't know if they will succeed or not. If we were very strict about [serializability](https://github.com/osmosis-labs/mesh-security/blob/main/docs/ibc/Serializability.md) in this case, we wouldn't allow any reward distribution until all in-flight packets (from all users) and finalized. This is unrealistic, so we can relax this a bit to "read committed", which is close enough as there are no invariants to enforce here, and we already accepted an approximation above.
>
> This means, that we don't update a user's reward shares for any in-flight IBC transaction until we have received a success ack. This is such an edge case on which side of the epoch they fall, and both decisions are equally valid from an economic perspective. However, in order to favor the approach of database consistency, we will use the "read committed" approach, and only count rewards after a successful ACK.

ethanfrey: Locking stake alone will avoid any conflicts (for the same user).
They will miss out of potential rewards but any other approach makes reward distribution unusable

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +98,122 @@ impl ExternalStakingContract<'_> {
...
+            } => (amount, user, validator),
+            _ => unreachable!(),
+        };
+
+        // Load stake
+        let mut stake_lock = self
+            .stakes
+            .load(ctx.deps.storage, (&tx_user, &tx_validator))?;
+
+        // Load distribution
```
ethanfrey, line 192: Same comment as above about not locking this, just stake

contracts/provider/external-staking/src/contract.rs
```diff
@@ -101,49 +231,171 @@ impl ExternalStakingContract<'_> {
...
 
-        stake.stake -= amount.amount;
+        stake_lock.lock_write()?;
+        self.stakes.save(
+            ctx.deps.storage,
+            (&ctx.info.sender, &validator),
+            &stake_lock,
+        )?;
+
+        let mut distribution_lock = self.distribution.load(ctx.deps.storage, &validator)?;
```
ethanfrey, line 252: Again, no lock here

contracts/provider/external-staking/src/contract.rs
```diff
@@ -101,49 +231,171 @@ impl ExternalStakingContract<'_> {
...
+        // Save tx
+        let new_tx = Tx::InFlightRemoteUnstaking {
+            id: tx_id,
+            amount: amount.amount,
+            user: ctx.info.sender.clone(),
+            validator,
+        };
+        self.pending_txs.save(ctx.deps.storage, tx_id, &new_tx)?;
+
+        // TODO: Remote IBC should happen here.
```
ethanfrey, line 270: Right here, you can trigger an ibc packet.

If we set multi-test up properly, it will totally ignore IBC messages (no action). This means we could use the IBC-enabled format here to send into `/dev/null` and then use the SudoMsg trick to trigger a fake ack. Which means most logic would be covered by current multi-test

contracts/provider/external-staking/src/contract.rs
```diff
@@ -101,49 +231,171 @@ impl ExternalStakingContract<'_> {
...
+
+        // Commit amount (need to unlock it first)
+        stake_lock.unlock_write()?;
+        let stake = stake_lock.write()?;
+        stake.stake -= tx_amount;
+
+        // FIXME? Release period being computed after successful IBC tx
         let release_at = ctx.env.block.time.plus_seconds(config.unbonding_period);
         let unbond = PendingUnbond {
-            amount: amount.amount,
```
ethanfrey, line 120: Good to do all this logic in commit, as it would be very tricky to roll this back.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +98,122 @@ impl ExternalStakingContract<'_> {
...
+            .save(ctx.deps.storage, (&tx_user, &tx_validator), &stake_lock)?;
+
+        // Save distribution
+        self.distribution
+            .save(ctx.deps.storage, &tx_validator, &distribution_lock)?;
+
+        // Remove tx
+        self.pending_txs.remove(ctx.deps.storage, tx_id);
+
+        // TODO: Call commit hook on vault
```
ethanfrey, line 158: Yes, I was just going to point this out.
A normal message should work here.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +98,122 @@ impl ExternalStakingContract<'_> {
...
+        distribution_lock.unlock_write()?;
+
+        // Save distribution
+        self.distribution
+            .save(ctx.deps.storage, &tx_validator, &distribution_lock)?;
+
+        // Remove tx
+        self.pending_txs.remove(ctx.deps.storage, tx_id);
+
+        // TODO: Call rollback hook on vault
```
ethanfrey, line 212: Key point

ethanfrey (commented):
Looks quite good.

A few missing TODOs I commented on, and some minor code nits.

Main point is removing lock on distribution.
And then exposing these commit/rollback methods as SudoMsg for testing.

If you can fix this up, I can work on the corresponding side for converter (and write the external-staking ack handlers as well if you want... just pull out the logic for an easy call)

contracts/provider/external-staking/src/msg.rs
maurolacy: Agreed. 🙂

contracts/provider/external-staking/src/multitest.rs
maurolacy: Agreed. Will do.

packages/sync/src/txs.rs
maurolacy: Yes, simplifying this to `InFlightRemoteStaking {{ id: {}}}` makes sense.

contracts/provider/external-staking/src/contract.rs
maurolacy: Good catch.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -100,118 +100,134 @@ impl ExternalStakingContract<'_> {
...
-        let stake = stake_lock.write()?;
-        stake.stake += tx_amount;
+    #[msg(exec)]
+    fn test_commit_stake(&self, ctx: ExecCtx, tx_id: u64) -> Result<Response, ContractError> {
+        #[cfg(test)]
+        {
+            // Load tx
+            let tx = self.pending_txs.load(ctx.deps.storage, tx_id)?;
+
+            // TODO: Verify tx comes from the right context
```
maurolacy, line 118: I think this TODO can go now.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +106,129 @@ impl ExternalStakingContract<'_> {
...
+                .distribution
+                .may_load(ctx.deps.storage, &tx_validator)?
+                .unwrap_or_default();
+
+            // Commit amount (need to unlock it first)
+            stake_lock.unlock_write()?;
+            let stake = stake_lock.write()?;
+            stake.stake += tx_amount;
+
+            // Commit distribution (need to unlock it first)
```
maurolacy, line 151: Remove outdated comment.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +106,129 @@ impl ExternalStakingContract<'_> {
...
+    /// Rollbacks a pending stake.
+    /// Must be called by the IBC callback handler on failed remote staking.
+    #[msg(exec)]
+    fn test_rollback_stake(&self, ctx: ExecCtx, tx_id: u64) -> Result<Response, ContractError> {
+        #[cfg(test)]
+        {
+            // Load tx
+            let tx = self.pending_txs.load(ctx.deps.storage, tx_id)?;
+
+            // TODO: Verify tx comes from the right context
```
maurolacy, line 189: Remove outdated TODO.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -101,51 +246,191 @@ impl ExternalStakingContract<'_> {
...
+            let msg = IbcMsg::SendPacket {
+                channel_id: channel.endpoint.channel_id,
+                data: to_binary(&packet)?,
+                timeout: packet_timeout(&ctx.env),
+            };
+            resp = resp.add_message(msg);
+        }
+
+        // put this later so compiler doens't complain about mut in test mode
+        resp = resp.add_attribute("owner", ctx.info.sender);
```
maurolacy, line 301: Nit: Move all the `add_attribute` calls here.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -84,6 +106,129 @@ impl ExternalStakingContract<'_> {
...
+            let _ = (ctx, tx_id);
+            Err(ContractError::Unauthorized {})
+        }
+    }
+
+    /// Rollbacks a pending stake.
+    /// Must be called by the IBC callback handler on failed remote staking.
+    #[msg(exec)]
+    fn test_rollback_stake(&self, ctx: ExecCtx, tx_id: u64) -> Result<Response, ContractError> {
+        #[cfg(test)]
```
maurolacy, line 184: The pattern in `converter`, in which you call an internal method here for tests and do nothing otherwise is better / clearer. I'll think you'll refactor to that, when connecting with the IBC handlers.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -101,51 +246,191 @@ impl ExternalStakingContract<'_> {
...
+    #[msg(exec)]
+    fn test_commit_unstake(&self, ctx: ExecCtx, tx_id: u64) -> Result<Response, ContractError> {
+        #[cfg(test)]
+        {
+            use crate::state::PendingUnbond;
+
+            // Load tx
+            let tx = self.pending_txs.load(ctx.deps.storage, tx_id)?;
+
+            // TODO: Verify tx comes from the right context
```
maurolacy, line 317: This is outdated.

maurolacy (commented):
Some comments on your recent changes.

contracts/provider/external-staking/src/contract.rs
ethanfrey: Yup, I just refactored to that, working on one step at a time to keep tests passing.

PR #62, opened 2023-06-20
Implement IBC channel handshake

ethanfrey:
- [x] Add version handshake logic
- [x] Extends documentation on channel handshake protocol 
- [x] Add all ibc entry points as stubs for external-staking and converter
- [x] Implement the 3 channel lifetime calls 
- [ ] Add E2E tests (with Go ibctesting?)


ethanfrey:
This is implemented and documented. We need E2E tests, but I think that is not blocking the merge

packages/apis/src/ibc/version.rs
```diff
@@ -0,0 +1,246 @@
...
+            })
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn to_string(&self) -> StdResult<String> {
+        let bytes = to_vec(self)?;
+        Ok(String::from_utf8(bytes)?)
+    }
```
maurolacy, line 113: Weird, this works. Will it in all platforms?

docs/ibc/ControlChannel.md
```diff
@@ -70,6 +70,20 @@ The general process (assuming a vault is already established on the provider) is
 3. Create IBC channel from provider to consumer
 4. Apply to consumer governance to provide a virtual staking max cap to the associated virtual staking contract, so that this connection may have voting power.
 
+### Handshake
```
maurolacy, line 73: 👍🏼 Good description.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -73,6 +74,10 @@ impl ExternalStakingContract<'_> {
 
         set_contract_version(ctx.deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
 
+        // Question: do we need to validate here? Is there a defined schema for such
+        // Question: should we add a query endpoint for this?
```
maurolacy, line 78: I think so. Just turn into a TODO.

maurolacy (approved):
LGTM.

packages/apis/src/ibc/version.rs
ethanfrey: What do you mean?

`String::from_utf8()` will error if the bytes are not valid utf-8.
The JSON encoding `to_vec()` will produce utf-8 encoded JSON as a `Vec<u8>`

If fact, the serde library implements it with an unchecked_utf8 conversion.... https://github.com/CosmWasm/serde-json-wasm/blob/main/src/ser/mod.rs#L433-L451

contracts/provider/external-staking/src/contract.rs
ethanfrey: Done in https://github.com/osmosis-labs/mesh-security/pull/62/commits/70a71ff4854d3208553b50c5d2ae8c2c1800521f

packages/apis/src/ibc/version.rs
maurolacy: Oh, I see. Thought `to_vec` was a generic / raw "struct to bytes" function. Got confused by the `bytes` var I guess.

PR #63, opened 2023-06-21
Implement IBC Validator Sync sub-protocol

ethanfrey:
Send the validator sync packets from Consumer (`converter`) to Provider (`external-staking`) and process them into the CRDT type there.

Add tests on the CRDT and some queries, so we can (in a future PR) check validators before staking, or associate tendermint double-signs with valoper addresses (next milestone)

### DIFF
--- a/contracts/consumer/converter/src/ibc.rs
+++ b/contracts/consumer/converter/src/ibc.rs
@@ -2,13 +2,15 @@
 use cosmwasm_std::entry_point;
 
 use cosmwasm_std::{
-    from_slice, DepsMut, Env, Ibc3ChannelOpenResponse, IbcBasicResponse, IbcChannel,
-    IbcChannelCloseMsg, IbcChannelConnectMsg, IbcChannelOpenMsg, IbcChannelOpenResponse,
-    IbcPacketAckMsg, IbcPacketReceiveMsg, IbcPacketTimeoutMsg, IbcReceiveResponse,
+    from_slice, to_binary, DepsMut, Env, Ibc3ChannelOpenResponse, IbcBasicResponse, IbcChannel,
+    IbcChannelCloseMsg, IbcChannelConnectMsg, IbcChannelOpenMsg, IbcChannelOpenResponse, IbcMsg,
+    IbcPacketAckMsg, IbcPacketReceiveMsg, IbcPacketTimeoutMsg, IbcReceiveResponse, IbcTimeout,
 };
 use cw_storage_plus::Item;
 
-use mesh_apis::ibc::{validate_channel_order, ProtocolVersion, PROTOCOL_NAME};
+use mesh_apis::ibc::{
+    validate_channel_order, AddValidator, ConsumerPacket, ProtocolVersion, PROTOCOL_NAME,
+};
 
 use crate::error::ContractError;
 
@@ -19,6 +21,16 @@ const MIN_IBC_PROTOCOL_VERSION: &str = "1.0.0";
 
 // IBC specific state
 const IBC_CHANNEL: Item<IbcChannel> = Item::new("ibc_channel");
+
+// Let those validator syncs take a day...
+const DEFAULT_TIMEOUT: u64 = 24 * 60 * 60;
+
+fn packet_timeout(env: &Env) -> IbcTimeout {
+    // No idea about their blocktime, but 24 hours ahead of our view of the clock
+    // should be decently in the future.
+    let timeout = env.block.time.plus_seconds(DEFAULT_TIMEOUT);
+    IbcTimeout::with_timestamp(timeout)
+}
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 /// enforces ordering and versioning constraints
@@ -63,7 +75,7 @@ pub fn ibc_channel_open(
 /// once it's established, we store data
 pub fn ibc_channel_connect(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     msg: IbcChannelConnectMsg,
 ) -> Result<IbcBasicResponse, ContractError> {
     // ensure we have no channel yet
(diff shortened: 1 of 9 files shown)


maurolacy (approved):
LGTM.

contracts/provider/external-staking/src/crdt.rs
```diff
@@ -0,0 +1,86 @@
...
+#[cw_serde]
+pub enum ValidatorState {
+    Active(ActiveState),
+    Tombstoned {},
+}
+
+#[cw_serde]
+/// Active state maintains a sorted list of updates with no duplicates.
+/// The first one is the one with the highest start_height.
+pub struct ActiveState(Vec<ValUpdate>);
```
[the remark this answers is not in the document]
ethanfrey: But BTreeSet is not good for `cw_serde` (I think).
I could use it internally

contracts/consumer/converter/src/ibc.rs
```diff
@@ -117,21 +148,28 @@ pub fn ibc_packet_receive(
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-/// never should be called as we do not send packets
+/// We get acks on sync state without much to do.
+/// If it succeeded, take no action. If it errored, we can't do anything else and let it go.
```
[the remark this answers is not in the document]
ethanfrey: Yes, maybe we log something with a custom event?
I will add that so there is some possibility of client-side handling

packages/apis/src/ibc/packet.rs
```diff
@@ -5,7 +5,7 @@ use cosmwasm_std::Coin;
 /// ibc_packet_receive in converter must handle them all.
 /// Each one has a different ack to be used in the reply.
 #[cw_serde]
-pub enum ProviderMsg {
+pub enum ProviderPacket {
```
maurolacy, line 8: After a second look with a fresh head: We would need to send the transaction id in these packets. So that it can be returned in the ack, to be committed.

packages/apis/src/ibc/packet.rs
```diff
@@ -24,58 +24,52 @@ pub enum ProviderMsg {
     },
 }
 
-/// Ack sent for ProviderMsg::Stake
+/// Ack sent for ProviderPacket::Stake
 #[cw_serde]
 pub struct StakeAck {}
```
maurolacy, line 29: This would need to include the tx id to be acknowledged.

In the same way, I think we'll need a series of "negative ack" (`Nack`) packets, also including the transaction id to be rolled back.

Without that, it would be impossible to track different in flight transactions, involving different operations / users.

maurolacy (commented):
Some more comments on IBC packets design / fields.

contracts/consumer/converter/src/ibc.rs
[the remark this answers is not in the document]
ethanfrey: Event added.
And this made me realize I wasn't using proper app wrapper that shows success or error

packages/apis/src/ibc/packet.rs
ethanfrey: Yes, I agree. The Provider side will need that. I was thinking that last night, and was gonna add that in a separate PR, but will do that here.

packages/apis/src/ibc/packet.rs
ethanfrey: When we get the ack callback, we also get the original packet (and the tx_id there). No need to include it in the ack itself. But I can if you want.

I just added the AckWrapper for the standard one done in ics20 and wasmd on contract errors. This handles success and error cases (which you are very correct to point out we do not handle).

Glad to see you pointing out these issues as well.

packages/apis/src/ibc/packet.rs
```diff
@@ -73,3 +75,25 @@ pub struct AddValidatorsAck {}
...
+
+/// This is a generic ICS acknowledgement format.
+/// Proto defined here: https://github.com/cosmos/cosmos-sdk/blob/v0.42.0/proto/ibc/core/channel/v1/channel.proto#L141-L147
+/// This is compatible with the JSON serialization.
+/// Wasmd uses this same wrapper for unhandled errors.
+#[cw_serde]
+pub enum AckWrapper {
+    Result(Binary),
+    Error(String),
+}
```
maurolacy: 👍🏼 Makes sense.

contracts/provider/external-staking/src/ibc.rs
```diff
@@ -132,13 +132,13 @@ pub fn ibc_packet_receive(
                 };
                 VAL_CRDT.add_validator(deps.storage, &valoper, update)?;
```
maurolacy, line 133: Do we capture / map the error(s) here, and return `ack_fail` instead?

packages/apis/src/ibc/packet.rs
ethanfrey: Added here: https://github.com/osmosis-labs/mesh-security/pull/63/commits/4fa8aecba56080bc3a96db7657406d612d0e3126

contracts/consumer/converter/src/ibc.rs
```diff
@@ -150,12 +152,27 @@ pub fn ibc_packet_receive(
...
 pub fn ibc_packet_ack(
     _deps: DepsMut,
     _env: Env,
-    _msg: IbcPacketAckMsg,
+    msg: IbcPacketAckMsg,
 ) -> Result<IbcBasicResponse, ContractError> {
-    Ok(IbcBasicResponse::new())
+    let ack: AckWrapper = from_slice(&msg.acknowledgement.data)?;
+    let mut res = IbcBasicResponse::new();
+    match ack {
```
maurolacy, line 163: 👍🏼 Taking shape.

packages/apis/src/ibc/packet.rs
ethanfrey: Done in https://github.com/osmosis-labs/mesh-security/pull/63/commits/4fa8aecba56080bc3a96db7657406d612d0e3126

contracts/provider/external-staking/src/crdt.rs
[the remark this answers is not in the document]
maurolacy: You can deserialise to it, and serialise it to `Vec`. It would be equivalent in the end to the current impl, so, I wouldn't bother.

It would be nicer if we can use it directly as a back storage.

contracts/provider/external-staking/src/ibc.rs
ethanfrey: This is done in wasmd. 

If contract panics, it aborts the whole tx (no ack written)
If contract returns error, it wraps the error message in this "standard" struct, which is what I use now as AckWrapper.

packages/apis/src/ibc/packet.rs
ethanfrey: Struct copied from cw20-ics20 (and methods updated).
This is tested to work with wasmd handling

PR #64, opened 2023-06-22
Add ibc handling to converter

ethanfrey:
(Currently untested)

- [x] Add ibc_packet_receive with price normalization and calling the virtual staking contract.
- [x] Add remote denom to the instantiate entry point
- [x] Solve converter / virtual_staking "chicken and egg" problem (see [ConverterContract::instantiate](https://github.com/osmosis-labs/mesh-security/blob/consumer-side-ibc-staking/contracts/consumer/converter/src/contract.rs#L36-L42))
- [x] Add ~~sudo entry point~~ test ExecMsg variants to allow testing
- [x] Write simple multi-test to test this

### DIFF
--- a/contracts/consumer/converter/src/contract.rs
+++ b/contracts/consumer/converter/src/contract.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{Addr, Decimal, Response};
+use cosmwasm_std::{to_binary, Addr, Coin, Decimal, Deps, DepsMut, Event, Response, WasmMsg};
 use cw2::set_contract_version;
 use cw_storage_plus::Item;
 use cw_utils::nonpayable;
@@ -6,6 +6,8 @@ use sylvia::types::{ExecCtx, InstantiateCtx, QueryCtx};
 use sylvia::{contract, schemars};
 
 use mesh_apis::converter_api::{self, ConverterApi, RewardInfo};
+use mesh_apis::price_feed_api;
+use mesh_apis::virtual_staking_api;
 
 use crate::error::ContractError;
 use crate::msg::ConfigResponse;
@@ -51,6 +53,7 @@ impl ConverterContract<'_> {
             price_feed: ctx.deps.api.addr_validate(&price_feed)?,
             // TODO: better error if discount greater than 1 (this will panic)
             adjustment: Decimal::one() - discount,
+            local_denom: ctx.deps.querier.query_bonded_denom()?,
         };
         self.config.save(ctx.deps.storage, &config)?;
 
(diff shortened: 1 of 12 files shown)


ethanfrey:
Okay, the consumer side is implemented.

Please review, but this has some tests and should be ready to merge.
I tried doing things like:

```rust
    #[cfg(test)]
    #[msg(exec)]
    fn demo_stake(
```

and

```rust
    #[cfg_attr(test, msg(exec))]
    fn demo_stake(
```

but neither worked, as `#[msg]` doesn't seem to be an attribute but is actually parsed strings by the `#[contract]` macro.
If you find/know a better way to conditionally compile these methods, only for testing, let me know.

I am happy to use a feature flag as well, but the main part is some sort of `cfg` directive that plays well with the `msg(exec)` directive

contracts/consumer/converter/src/ibc.rs
```diff
@@ -142,11 +142,41 @@ pub fn ibc_channel_close(
...
 pub fn ibc_packet_receive(
-    _deps: DepsMut,
+    deps: DepsMut,
     _env: Env,
-    _msg: IbcPacketReceiveMsg,
+    msg: IbcPacketReceiveMsg,
 ) -> Result<IbcReceiveResponse, ContractError> {
-    todo!();
+    let packet: ProviderPacket = from_slice(&msg.packet.data)?;
+    let contract = ConverterContract::new();
```
maurolacy, line 150: Interesting. `sylvia`'s interfaces at work.

contracts/consumer/converter/src/contract.rs
```diff
@@ -44,24 +55,87 @@ impl ConverterContract<'_> {
...
+    /// Ideally we want conditional compilation of these whole methods and the enum variants
+    #[msg(exec)]
+    fn demo_stake(
+        &self,
+        ctx: ExecCtx,
+        validator: String,
+        stake: Coin,
+    ) -> Result<Response, ContractError> {
+        // This can only ever be valid in tests
+        ensure_eq!(ctx.info.sender, "ADMIN", ContractError::Unauthorized);
```
maurolacy, line 121: Was thinking of doing the same actually. Will give the `sudo` alternative a try asap.

contracts/consumer/converter/src/contract.rs
```diff
@@ -44,24 +55,87 @@ impl ConverterContract<'_> {
...
+        let init_data = parse_instantiate_response_data(&reply.data.unwrap())?;
+        let virtual_staking = Addr::unchecked(init_data.contract_address);
+        self.virtual_stake.save(deps.storage, &virtual_staking)?;
         Ok(Response::new())
     }
 
+    /// This is only used for tests
+    /// Ideally we want conditional compilation of these whole methods and the enum variants
+    #[msg(exec)]
+    fn demo_stake(
```
maurolacy, line 114: Nit: perhaps `test_stake` or `stake_test` is a better name.

contracts/consumer/converter/src/contract.rs
```diff
@@ -72,6 +146,79 @@ impl ConverterContract<'_> {
...
+        validator: String,
+        stake: Coin,
+    ) -> Result<Response, ContractError> {
+        let amount = self.normalize_price(deps.as_ref(), stake)?;
+
+        let event = Event::new("mesh-bond")
+            .add_attribute("validator", &validator)
+            .add_attribute("amount", amount.amount.to_string());
+
+        let msg = virtual_staking_api::ExecMsg::Bond { validator, amount };
```
maurolacy, line 164: Nit: Shouldn't this better be called `Delegate`?

contracts/consumer/converter/src/contract.rs
```diff
@@ -72,6 +146,79 @@ impl ConverterContract<'_> {
...
+            contract_addr: self.virtual_stake.load(deps.storage)?.into(),
+            msg: to_binary(&msg)?,
+            funds: vec![],
+        };
+
+        Ok(Response::new().add_message(msg).add_event(event))
+    }
+
+    fn normalize_price(&self, deps: Deps, amount: Coin) -> Result<Coin, ContractError> {
+        // TODO: ensure the proper remote denom - set this in the instantiate
```
maurolacy, line 199: Remove this TODO

contracts/consumer/converter/src/contract.rs
```diff
@@ -72,6 +146,79 @@ impl ConverterContract<'_> {
...
+                sent: amount.denom,
+                expected: config.remote_denom
+            }
+        );
+
+        // get the price value (usage is a bit clunky, need use trait and cannot chain Remote::new() with .querier())
+        use price_feed_api::Querier;
+        let remote = price_feed_api::Remote::new(config.price_feed);
+        let price_feed = remote.querier(&deps.querier);
+        let price = price_feed.price()?.native_per_foreign;
```
maurolacy, line 214: Guess this query will look something like

`let price = price_feed.price(config.remote_denom)?.native_per_foreign;`

in the (near?) future.

contracts/consumer/converter/src/contract.rs
```diff
@@ -72,6 +146,79 @@ impl ConverterContract<'_> {
...
+        let converted = (amount.amount * price) * config.adjustment;
```
maurolacy, line 215: Nit: Rename `adjustment` to `price_adjustment`.

maurolacy (approved):
Looks good! Will take a look at the tests next.

maurolacy:
> neither worked, as `#[msg]` doesn't seem to be an attribute but is actually parsed strings by the `#[contract]` macro. If you find/know a better way to conditionally compile these methods, only for testing, let me know.
> 
> I am happy to use a feature flag as well, but the main part is some sort of `cfg` directive that plays well with the `msg(exec)` directive

Was trying the same to no avail. I think we need to use some form of `sudo`. The trick with the "ADMIN" check is something I thought but decided not to do. Will work on improving this asap.

contracts/consumer/converter/src/ibc.rs
ethanfrey: Yeah, how I did in sudo. I put the (unsupported) entry point outside of the struct and then call into methods.

contracts/consumer/converter/src/contract.rs
ethanfrey: Good question, can you make a different issue on this?
Out of scope of the IBC connection

contracts/consumer/converter/src/contract.rs
ethanfrey: Yes, related issue: https://github.com/CosmWasm/sylvia/issues/181
I will add this to code comment

ethanfrey:
BTW, I settled on this:

```rust
#[msg(exec)]
    fn test_stake(
        &self,
        ctx: ExecCtx,
        validator: String,
        stake: Coin,
        #[cfg(test)] {
            // This can only ever be called in tests
            self.stake(ctx.deps, validator, stake)
        }
        #[cfg(not(test))] {
            let _ = (ctx, validator, stake);
            Err(ContractError::Unauthorized)
        }
    }
```

The entry point is still there, but it is compile-time fixed to errors in non-test builds.

PR #66, opened 2023-06-26
Fix details

maurolacy:
A small PR with some comments updates / style fixes after a quick review of the main recent changes.


contracts/provider/external-staking/src/contract.rs
```diff
@@ -844,8 +844,11 @@ pub mod cross_staking {
                 resp = resp.add_message(msg);
             }
 
-            // This is later so `mut resp` is valid in test and non-test code
-            resp = resp.add_attribute("tx_id", tx_id.to_string());
+            resp = resp
```
ethanfrey, line 847: Ah, it does look better there

ethanfrey (approved):
Nice fixes

PR #73, opened 2023-06-27
Better ibc tests

ethanfrey:
Try to accept (ignore) IbcMsg so we can run the full tests, but I cannot customize sylvia mt wrapper sufficiently

### DIFF
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -704,14 +704,14 @@ fn distribution() {
     // 20 tokens for users[0]
     // 30 tokens for users[1]
     contract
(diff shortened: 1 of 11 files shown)


contracts/provider/external-staking/src/multitest.rs
```diff
@@ -66,7 +66,10 @@ fn setup<'app>(
 
 #[test]
 fn instantiate() {
-    let app = App::default();
+    let mt_app = cw_multi_test::AppBuilder::new()
```
ethanfrey, line 69: This errors as sylvia::multitest_utils::App doesn't let me override the module types

JakeHartnell:
Closing as stale, seems like most of this already made it in.

PR #74, opened 2023-06-27
Ensure validators are registered before staking

### LINKED ISSUE #70: Reject unknown validators for external staking
I was testing with a provider chain validator address that got accepted and eventually failed in epoch rebalance on the consumer side. This should fail faster

ethanfrey:
Closes #70 

This doesn't let a new stake occur unless when know the validator is in the active set.
Added a test command to place a validator in the active set.

### DIFF
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -1,6 +1,7 @@
 use anyhow::Result as AnyResult;
 
 use cosmwasm_std::{coin, coins, to_binary, Addr, Decimal};
+use mesh_apis::ibc::AddValidator;
 use mesh_native_staking::contract::multitest_utils::CodeId as NativeStakingCodeId;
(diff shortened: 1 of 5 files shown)


contracts/provider/external-staking/src/multitest.rs
```diff
@@ -115,6 +125,23 @@ fn staking() {
         .call(users[1])
         .unwrap();
 
+    /*
```
ethanfrey, line 128: @hashedone @jawoznia 

Another non-blocking issue in Sylvia.

User calls contract A. contract A returns message to execute contract B. contract B calls fail (no reply registered). A and B both have different ContractError registered.

The caller doesn't ever get the result, but it fails inside ExecProxy when trying to downcast the result to vault::ContractError (contract A). Not sure if the error is external_staking::ContractError (contract B), or StdError, but I want to assert this fails without panic-ing the test

ethanfrey:
Merging this one. Happy for more detailed review later.

PR #78, opened 2023-06-27
Configure max slashable

### LINKED ISSUE #71: All configuring the max_slashable amount for staking
Currently, for MVP, it is [hardcoded at 10% for local staking](https://github.com/osmosis-labs/mesh-security/blob/v0.3.0-beta.2/contracts/provider/native-staking/src/contract.rs#L21-L22) and set to [5% for remote staking](https://github.com/osmosis-labs/mesh-security/blob/v0.3.0-beta.2/contracts/provider/external-staking/src/contract.rs#L857-L867)

We could easily add an argument for this to the InstantiateMsg of both contracts. Maybe there is some way to query that directly from the chain, but if not, let's just make it optional on instantiate.

ethanfrey:
Closes #71 

Allows configuring max_slashing for native-staking and external-staking, making them more versatile and especially providing better test cases (when this is high, the total slashable becomes a factor).

Note this adds a new parameter to the InstantiateMsg of both of those contracts `max_slashable: Decimal`

### DIFF
--- a/contracts/provider/external-staking/src/multitest.rs
+++ b/contracts/provider/external-staking/src/multitest.rs
@@ -21,6 +21,8 @@ use crate::msg::{AuthorizedEndpoint, ReceiveVirtualStake, StakeInfo};
 const OSMO: &str = "osmo";
 const STAR: &str = "star";
 
+const SLASHING_PERCENTAGE: u64 = 10;
+
 // Shortcut setuping all needed contracts
 //
 // Returns vault and external staking proxies
@@ -59,6 +61,7 @@ fn setup<'app>(
             vault.contract_addr.to_string(),
             unbond_period,
             remote_contact,
+            Decimal::percent(SLASHING_PERCENTAGE),
         )
         .call(owner)?;
 
@@ -78,7 +81,7 @@ fn instantiate() {
     assert_eq!(stakes.stakes, []);
 
     let max_slash = contract.cross_staking_api_proxy().max_slash().unwrap();
-    assert_eq!(max_slash.max_slash, Decimal::percent(5));
+    assert_eq!(max_slash.max_slash, Decimal::percent(SLASHING_PERCENTAGE));
 }
 
 #[test]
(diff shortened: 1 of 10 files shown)


contracts/provider/native-staking/src/local_staking_api.rs
```diff
@@ -79,9 +80,10 @@ impl LocalStakingApi for NativeStakingContract<'_> {
     /// Returns the maximum percentage that can be slashed
     /// TODO: Any way to query this from the chain? Or we just pass in InstantiateMsg?
```
maurolacy, line 81: Remove this TODO.

contracts/provider/external-staking/src/state.rs
```diff
@@ -13,8 +13,10 @@ pub struct Config {
     pub rewards_denom: String,
     /// Vault contract address
     pub vault: VaultApiHelper,
-    /// Ubbounding period for claims in seconds
+    /// Unbounding period for claims in seconds
```
maurolacy, line 16: Unbounding -> Unbonding. 🙂

contracts/provider/external-staking/src/multitest.rs
```diff
@@ -37,6 +39,7 @@ fn setup<'app>(
     let native_staking_instantiate = NativeStakingInstantiateMsg {
         denom: OSMO.to_owned(),
         proxy_code_id: native_staking_proxy_code.code_id(),
+        max_slashing: Decimal::percent(5),
```
maurolacy, line 42: Is this to distinguish local from external slashing percentages? (i.e. they don't need to be the same). Perhaps use two different constants then?

maurolacy (approved):
LGTM.

contracts/provider/external-staking/src/multitest.rs
ethanfrey: Yeah, I just wanted to set up different values, so if there are future tests on slashing it is clear which value is used. I will add a constant

PR #79, opened 2023-06-27
List all rewards per user

### LINKED ISSUE #47: List of pending rewards per user in external staking
Bonus: it would be nice to have a list of all pending rewards for a user.

I guess you could use `self.stakes.prefix(&user).range(...)` to get that list and return caluclate_reward for each one.

Even without pagination, it would be very useful for any UI

_Originally posted by @ethanfrey in https://github.com/osmosis-labs/mesh-security/pull/46#discussion_r1217896481_

ethanfrey:
Closes #47 

Helpful query to view all rewards a user has over all validators when cross-staking

### DIFF
--- a/contracts/provider/external-staking/src/contract.rs
+++ b/contracts/provider/external-staking/src/contract.rs
@@ -1,6 +1,6 @@
 use cosmwasm_std::{
     coin, coins, ensure, ensure_eq, from_binary, Addr, BankMsg, Binary, Coin, Decimal, DepsMut,
-    Env, Order, Response, StdResult, Storage, Uint128, Uint256, WasmMsg,
+    Env, Order, Response, StdError, StdResult, Storage, Uint128, Uint256, WasmMsg,
 };
 use cw2::set_contract_version;
 use cw_storage_plus::{Bounder, Item, Map};
@@ -25,9 +25,9 @@ use sylvia::types::{ExecCtx, InstantiateCtx, QueryCtx};
 
 use crate::error::ContractError;
 use crate::msg::{
-    AllTxsResponse, AuthorizedEndpointResponse, ConfigResponse, IbcChannelResponse,
-    ListRemoteValidatorsResponse, PendingRewards, ReceiveVirtualStake, StakeInfo, StakesResponse,
-    TxResponse,
+    AllPendingRewards, AllTxsResponse, AuthorizedEndpointResponse, ConfigResponse,
+    IbcChannelResponse, ListRemoteValidatorsResponse, PendingRewards, ReceiveVirtualStake,
+    StakeInfo, StakesResponse, TxResponse, ValidatorPendingReward,
 };
 use crate::state::{Config, Distribution, Stake};
 use mesh_sync::Tx;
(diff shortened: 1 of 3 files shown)


contracts/provider/external-staking/src/msg.rs
```diff
@@ -89,12 +89,33 @@ pub struct UsersResponse {
...
 }
 
-/// Response for penging rewards query
+/// Response for pending rewards query on one validator
 #[cw_serde]
 pub struct PendingRewards {
     pub amount: Coin,
 }
 
+/// Response for pending rewards query on one validator
```
maurolacy, line 98:
```suggestion
/// Response for pending rewards query on all validators
```

contracts/provider/external-staking/src/msg.rs
```diff
@@ -89,12 +89,33 @@ pub struct UsersResponse {
...
+#[cw_serde]
+pub struct AllPendingRewards {
+    pub rewards: Vec<ValidatorPendingReward>,
+}
+
+#[cw_serde]
+pub struct ValidatorPendingReward {
+    pub amount: Coin,
+    pub validator: String,
```
maurolacy, line 107: Nit: Switch the order of these for clarity.

contracts/provider/external-staking/src/msg.rs
```diff
@@ -89,12 +89,33 @@ pub struct UsersResponse {
...
+}
+
+impl ValidatorPendingReward {
+    pub fn new(amount: u128, denom: impl Into<String>, validator: impl Into<String>) -> Self {
```
maurolacy, line 111: Here too, I would put `validator` first, and then `amount` and `denom`.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -793,17 +793,54 @@ impl ExternalStakingContract<'_> {
...
+            .take(limit)
+            .map(|item| {
+                let (validator, stake_lock) = item?;
+                let stake = stake_lock.read()?;
+                let distribution = self
+                    .distribution
+                    .may_load(ctx.deps.storage, &validator)?
+                    .unwrap_or_default();
+                let amount = Self::calculate_reward(stake, &distribution)?;
+                Ok::<_, ContractError>(ValidatorPendingReward {
```
maurolacy, line 826: Nit: use the `new` helper.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -793,17 +793,54 @@ impl ExternalStakingContract<'_> {
...
+        let config = self.config.load(ctx.deps.storage)?;
+
+        let rewards: Vec<_> = self
+            .stakes
+            .prefix(&user)
+            .range(ctx.deps.storage, bound, None, Order::Ascending)
+            .take(limit)
+            .map(|item| {
+                let (validator, stake_lock) = item?;
+                let stake = stake_lock.read()?;
```
maurolacy, line 820: Use a `MaybeReward` wrapper not to error here. I can do it in a follow-up PR.

maurolacy (approved):
LGTM.

contracts/provider/external-staking/src/contract.rs
ethanfrey: Yeah, seems like nice follow up. You should also update pending_rewards to do the same (that errors on lock as well).
I just made these two work the same, happy for follow up to improve them both (and a test for querying with in-flight tx)

maurolacy: See #86.

PR #107, opened 2023-08-17
F/value range add sub helpers

maurolacy:
Add a couple of helpers to `ValueRange` for non-transactional addition and subtraction.

~~Names clash sometimes (but not always) with the `std::ops::Add` impl. That's unfortunate, but I couldn't think of better names. In any case, clashing can be avoided by using the fully qualified syntax for addition~~. (solved it by replacing the `Add` impl by a helper function)

Also adds a stock `Display` impl.

### DIFF
--- a/packages/sync/src/range.rs
+++ b/packages/sync/src/range.rs
@@ -83,7 +83,7 @@ pub fn max_range<T: Ord + Copy>(a: ValueRange<T>, b: ValueRange<T>) -> ValueRang
 // TODO: deprecate this?
 /// Problem: We have a list of ValueRanges, and we want to know the maximum value.
 /// This is not one clear value, as we consider the maximum if all commit and maximum if all rollback.
-/// The result is the range of possible maximum values (different than spread)  
+/// The result is the range of possible maximum values (different than spread)
 pub fn reduce_max_range<'a, I, T>(iter: I) -> ValueRange<T>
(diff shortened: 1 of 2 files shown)


packages/sync/src/range.rs
```diff
@@ -210,6 +211,16 @@ where
         self.assert_valid_range();
     }
 
+    /// This is a convenience method for non-transactional addition.
+    /// Similar to `prepare_add`, if the last value is set it enforces that the new maximum
+    /// will remain under that limit.
+    /// Usage: `range.add(20, None)?;` or `range.add(20, 100)?;`
+    pub fn add(&mut self, value: T, max: impl Into<Option<T>>) -> Result<(), RangeError> {
```
jawoznia, line 218: As discussed offline please consider removing impl of `Add` and `Sub` as they are only used in tests and I don't see any scenario in which this could be used in production code.

jawoznia (commented):
Not necessary in this PR but we could change `prepare_add` to something like `reserve_add`. This keyword well describes the purpose and is also used in the comment.

Also should this be merged to the `main`?

packages/sync/src/range.rs
maurolacy: Done. It's used in tests, so I transformed it to a helper function.

PR #108, opened 2023-08-21
F/value ranges external staking

jawoznia:


Cargo.toml
```diff
@@ -3,56 +3,55 @@ members = ["packages/*", "contracts/provider/*", "contracts/consumer/*"]
 resolver = "2"
 
 [workspace.package]
-edition       = "2021"
```
maurolacy, line 6: Nit: Not a big deal, but can we avoid doing this?

contracts/provider/external-staking/src/msg.rs
```diff
@@ -80,15 +80,15 @@ impl MaybeStake {
 pub struct StakeInfo {
     pub owner: String,
     pub validator: String,
-    pub stake: MaybeStake,
+    pub stake: Stake,
```
maurolacy, line 83: Nit: `MaybeStake` can now be removed.

contracts/provider/external-staking/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod contract;
 pub mod crdt;
+pub mod cross_staking;
```
maurolacy, line 3: ? Why is this needed?

contracts/provider/external-staking/src/contract.rs
```diff
@@ -155,9 +161,10 @@ impl ExternalStakingContract<'_> {
             .unwrap_or_default();
 
         // Commit amount (need to unlock it first)
-        stake_lock.unlock_write()?;
-        let stake = stake_lock.write()?;
-        stake.stake += tx_amount;
+        // stake_lock.unlock_write()?;
+        // let stake = stake_lock.write()?;
+        // stake.stake += tx_amount;
+        stake.stake.commit_add(tx_amount);
```
maurolacy, line 167: Originally, this is a "late addition" during `commit`. Instead of an "early addition" in the `receive_virtual_stake` method (like the usages in the `vault` contract).

But since we now change it to an early addition, this is correct.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -208,15 +215,16 @@ impl ExternalStakingContract<'_> {
...
         };
 
+        // What is the point of this?
         // Load stake
-        let mut stake_lock = self.stakes.load(deps.storage, (&tx_user, &tx_validator))?;
+        let mut stake = self.stakes.load(deps.storage, (&tx_user, &tx_validator))?;
 
         // Release stake lock
-        stake_lock.unlock_write()?;
+        stake.stake.rollback_add(tx_amount);
```
maurolacy, line 223: This is correct (in thew new setup) as well. There's now need to  rollback, because the `release_virtual_stake` is not a "do nothing" method anymore.

contracts/provider/external-staking/src/cross_staking.rs
```diff
@@ -0,0 +1,109 @@
+use cosmwasm_std::{ensure_eq, from_binary, to_binary, Binary, Coin, IbcMsg, Response, Uint128};
```
maurolacy, line 1: Why did you refactored this into its own module? It makes it a little more difficult to track the diffs.

contracts/provider/external-staking/src/cross_staking.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+        let mut stake = self
+            .stakes
+            .may_load(ctx.deps.storage, (&owner, &msg.validator))?
+            .unwrap_or_default();
+
+        // Write lock and save stake and distribution
+        // stake.lock_write()?;
+        // TODO: Possibly it should be add
+        stake.stake.prepare_sub(amount.amount, Uint128::zero())?;
```
maurolacy, line 57: Originally, we just lock and create a transaction here, and do the changes in the associated `commit` method.

Now it should be the same, but without locking. Then, we then need to **add** (not subtract) the new stakes. This is because we cannot keep with the logic of "lock and do nothing", i.e. the "late addition" mechanism; because we want to inform the user of the pending staking, through the value range change.

maurolacy (commented):
Didn't review all. Publishing because there's an error in the cross staking module.

Cargo.toml
jawoznia: I will revert it once it will be ready to review. It was made by auto formatter

contracts/provider/external-staking/src/contract.rs
```diff
@@ -300,53 +296,51 @@ impl ExternalStakingContract<'_> {
...
             .stakes
-            .may_load(ctx.deps.storage, (&ctx.info.sender, &validator))?
+            .may_load(deps.storage, (&info.sender, &validator))?
             .unwrap_or_default();
-        let stake = stake_lock.read()?;
 
         ensure!(
-            stake.stake >= amount.amount,
-            ContractError::NotEnoughStake(stake.stake)
+            stake.stake.high() >= amount.amount,
```
maurolacy, line 316: This has to be the low value of the stake, not the high one. If not, you would be able to unstake more than possible, by staking first an amount that is then rolled back (for whatever reason).

Not sure how / if this can be exploited, but the safest bet is to check against the low stake here.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -412,9 +406,10 @@ impl ExternalStakingContract<'_> {
             .unwrap_or_default();
 
         // Commit amount (need to unlock it first)
-        stake_lock.unlock_write()?;
-        let stake = stake_lock.write()?;
-        stake.stake -= tx_amount;
+        stake.stake.commit_sub(tx_amount);
+        // stake_lock.unlock_write()?;
+        // let stake = stake_lock.write()?;
+        // stake.stake -= tx_amount;
```
maurolacy, line 412: Remove commented code.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -522,24 +517,21 @@ impl ExternalStakingContract<'_> {
...
             .range(ctx.deps.storage, None, None, Order::Ascending)
             .collect::<Result<_, _>>()?;
 
-        let released: Uint128 = stake_locks
+        let released: Uint128 = stakes
             .into_iter()
-            .map(|(validator, mut stake_lock)| -> Result<_, ContractError> {
-                let stake = stake_lock.write()?;
+            .map(|(validator, mut stake)| -> Result<_, ContractError> {
+                // Should we zero out the stake here?
```
maurolacy, line 529: No.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -669,12 +659,9 @@ impl ExternalStakingContract<'_> {
 
         // lock the stake. the withdrawn_funds will be updated on a commit,
         // left unchanged on rollback
```
maurolacy, line 661: Nit: Update comment accordingly.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -669,12 +659,9 @@ impl ExternalStakingContract<'_> {
...
-        stake_lock.lock_write()?;
-        self.stakes.save(
-            ctx.deps.storage,
-            (&ctx.info.sender, &validator),
-            &stake_lock,
-        )?;
+        stake.stake.prepare_add(amount, None)?;
```
maurolacy, line 662: This is wrong. Stake does not change during rewards withdrawal.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -766,20 +753,23 @@ impl ExternalStakingContract<'_> {
...
             _ => {
                 return Err(ContractError::WrongTypeTx(tx_id, tx));
             }
         };
 
         // release the write lock and leave state unchanged
-        let mut stake_lock = self.stakes.load(deps.storage, (&staker, &validator))?;
-        stake_lock.unlock_write()?;
+        let mut stake = self.stakes.load(deps.storage, (&staker, &validator))?;
+        stake.stake.rollback_add(tx_amount);
```
maurolacy, line 770: This is incorrect. As mentioned, stake is unaffected by rewards processing.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -1040,8 +1015,11 @@ impl ExternalStakingContract<'_> {
     // could be enforced by taking user and validator in arguments, then fetching data, but
     // sometimes data are used also for different calculations so we want to avoid double
     // fetching.
-    fn calculate_reward(stake: &Stake, distribution: &Distribution) -> Result<Uint128, StdError> {
-        let points = distribution.points_per_stake * Uint256::from(stake.stake);
+    fn calculate_reward(
+        stake: &Stake,
+        distribution: &Distribution,
+    ) -> Result<Uint128, ContractError> {
+        let points = distribution.points_per_stake * Uint256::from(stake.stake.low());
```
maurolacy, line 1022: This is a good decision. It might introduce small errors against the user during pending unstakes, but since they are small, and temporary, I consider this is acceptable.

A staking operation is considered complete (so that it counts for rewards) only after it has been committed.
A unstacking operation on the contrary, is considered processed as soon as it's initiated. This is a (small) error (for rewards computation), but good for simplicity.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -1087,15 +1069,15 @@ pub mod cross_staking {
...
             }
-            let mut stake_lock = self
+            let mut stake = self
                 .stakes
                 .may_load(ctx.deps.storage, (&owner, &msg.validator))?
                 .unwrap_or_default();
 
-            // Write lock and save stake and distribution
-            stake_lock.lock_write()?;
+            // Save stake
```
maurolacy, line 1077:
```suggestion
            // Prepare stake addition and save stake.
            // We don't check for max here, as this call can only come from the `vault` contract, which already
            // performed the proper check.
```

maurolacy (commented):
I think your code isn't ready yet as it's still in draft, but did a 2nd review pass just out of curiosity.

contracts/provider/vault/src/contract.rs
```diff
@@ -691,6 +690,17 @@ impl VaultContract<'_> {
 
         Ok(())
     }
+
+    #[msg(query)]
+    fn user_collateral(&self, ctx: QueryCtx, user: String) -> Result<Uint128, ContractError> {
```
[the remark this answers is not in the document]
jawoznia: Removed as it is redundant and unnecessary

contracts/provider/external-staking/src/contract.rs
```diff
@@ -466,14 +458,14 @@ impl ExternalStakingContract<'_> {
         };
 
         // Load stake
-        let mut stake_lock = self.stakes.load(deps.storage, (&tx_user, &tx_validator))?;
+        let mut stake = self.stakes.load(deps.storage, (&tx_user, &tx_validator))?;
 
         // Release stake lock
```
maurolacy, line 463: Nit: Update comment.

In general, all the mentions to "locking" should be removed / updated.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -667,15 +653,6 @@ impl ExternalStakingContract<'_> {
...
             .add_attribute("amount", amount.to_string());
 
-        // lock the stake. the withdrawn_funds will be updated on a commit,
-        // left unchanged on rollback
-        stake_lock.lock_write()?;
-        self.stakes.save(
-            ctx.deps.storage,
-            (&ctx.info.sender, &validator),
-            &stake_lock,
-        )?;
```
maurolacy, line 677: 👍🏼  `stake` is now immutable so no need to save here.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -761,26 +738,9 @@ impl ExternalStakingContract<'_> {
...
 
-        // Verify tx is of the right type and get data
-        let (staker, validator) = match tx {
-            Tx::InFlightTransferFunds {
-                staker, validator, ..
-            } => (staker, validator),
-            _ => {
-                return Err(ContractError::WrongTypeTx(tx_id, tx));
-            }
-        };
```
maurolacy, line 776: Hmmm, I would just load and check the tx type for safety. Only then delete the tx.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -761,26 +738,9 @@ impl ExternalStakingContract<'_> {
...
-
-        // release the write lock and leave state unchanged
-        let mut stake_lock = self.stakes.load(deps.storage, (&staker, &validator))?;
-        stake_lock.unlock_write()?;
-        self.stakes
-            .save(deps.storage, (&staker, &validator), &stake_lock)?;
```
maurolacy, line 782: 👍🏼  value ranges are nicer than locks.

maurolacy (commented):
Looks good.

Will take a look at tests later.

contracts/provider/external-staking/src/contract.rs
jawoznia: agree ;)

contracts/provider/external-staking/src/error.rs
```diff
@@ -1,7 +1,7 @@
 use cosmwasm_std::{ConversionOverflowError, StdError, Uint128};
 use cw_utils::PaymentError;
 use mesh_apis::ibc::VersionError;
-use mesh_sync::{LockError, Tx};
+use mesh_sync::{LockError, RangeError, Tx};
```
maurolacy, line 4: Nit: from `LockError` can be removed as well.

contracts/provider/external-staking/src/multitest.rs
```diff
@@ -466,27 +463,23 @@ fn unstaking() {
...
         .unwrap();
-    assert_eq!(stake.stake.u128(), 240);
+    assert_eq!(stake.stake, ValueRange::new_val(Uint128::new(240)));
 
     let stake = contract
         .stake(users[1].to_string(), validators[1].to_string())
-        .unwrap()
         .unwrap();
-    assert_eq!(stake.stake.u128(), 0);
+    assert_eq!(stake.stake, ValueRange::new_val(Uint128::new(0)));
```
maurolacy, line 482: Nit:
```suggestion
    assert_eq!(stake.stake, ValueRange::new_val(Uint128::zero()));
```

contracts/provider/external-staking/src/multitest.rs
```diff
@@ -710,6 +699,8 @@ fn distribution() {
         )
         .call(users[0])
         .unwrap();
+
+    println!("Stake finished");
```
maurolacy, line 703: Nit: These can be removed.

maurolacy (approved):
Reviewed tests. Looks good.

There's still a `Lock` error that can be removed. And a `MaybePendingRewards` enum that can be turned into a `PendingRerwards` struct. I can do this last one if you want.

contracts/provider/external-staking/src/contract.rs
maurolacy: Let's add a short comment about this in the code.

maurolacy:
Can be merged, but better if you review mine before merging this on top.

contracts/provider/external-staking/src/error.rs
maurolacy: `Lockable` is part of the `sync` package, so I would leave it, as it can be useful in the future, and doesn't add any overhead to contracts.

contracts/provider/external-staking/src/contract.rs
maurolacy: Was just thinking that it would be good to add this check (stake unaffected by rewards processing) to the tests.

PR #109, opened 2023-08-22
F/valset updates

### LINKED ISSUE #88: Send validator sync update messages on changing validator set
Currently we only sync the original validator set but do not send new messages when it changes. 

To do so, we need more callbacks from the SDK https://github.com/osmosis-labs/mesh-security-sdk/issues/47

Once that is complete, we need to adapt the contracts to it and propagate the proper ibc packets.

maurolacy:
Closes #88.

TODO:

- [x] Unit tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1023,9 +1023,9 @@ checksum = "6bdef32e8150c2a081110b42772ffe7d7c9032b606bc226c8260fd97e0976601"
 
 [[package]]
 name = "sylvia"
-version = "0.6.1"
+version = "0.7.1"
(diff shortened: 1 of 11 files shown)


JakeHartnell (approved):
LGTM!

JakeHartnell:
Merging, if you have comments, please submit in the form of PRs. : )

maurolacy:
What is still needed here, and for completeness, would be to add a `jails` field with a list of validators to be jailed, to the `ValsetUpdate` message.

Putting this here just for the record. This is irrelevant at this point, but it could be useful / needed if we implement #118. To differentiate between the `unbonded` state and the (temporarily) `jailed` state.

PR #110, opened 2023-08-23
tests: Remove unused `packages/mocks`

jawoznia:


maurolacy (approved):
LGTM. I don't think anybody is using these. Looks like an early design decision that we lost track of during development.

Issue #11, opened 2023-04-24
Implement Slashing


ethanfrey:
TODO:

- [x] Slashing design considerations / requirements (Check / define with Osmosis).
- [x] Slashing design.
- [ ] Slashing implementation.

Extra issues:

- [ ] Detecting Byzantine chains. Decide on proper handling of this case, and see how to get updates from the IBC modules to detect such a state. (From [Detecting Byzantine Chains](https://github.com/osmosis-labs/mesh-security/blob/main/docs/ibc/Slashing.md#detecting-byzantine-chains)).
- [ ] Slashing Propagation: Consider the effects of slashing changes on the consistency of the staking state between Provider and Consumer.
   - [ ] For slashing ratio increase (if allowed).
   - [ ] For slashing ratio decrease (if allowed).

jtremback:
https://github.com/cosmos/interchain-security/pull/1128/files#diff-655d4a7e003e101281bc9490676d87f4b44c177da997cc3caa48a0b75986bc21

maurolacy:
OK, after discussing about this ,I think we now have a good understanding of the requirements and design decisions around cross-slashing. Here are the salient points:

Control channel: We can hook into the evidence module to receive a callback on the Consumer side:

https://github.com/cosmos/cosmos-sdk/blob/v0.47.4/x/evidence/abci.go#L24

We then pass the evidence as part of the mesh-security protocol / channel. Otherwise we would need deep chain integration on the Provider side.
This builds upon the assumption that the control channel, that is, the Consumer implementation of mesh security can be fully trusted. I would say, as Providers we trust that, but want to verify the evidence ourselves.

For verifying the evidence on the Provider side, we would need the Consumer's validator set with the validator pubkeys, at the evidence height. That means / implies that we need to store or have access to the historical validator set. At least for the unbonding period, which is typically 21 days.
This means either storing the validator set updates as part of the control channel protocol, which is already in planning (https://github.com/osmosis-labs/mesh-security/issues/88) or accessing the historical validator set or part of it on demand.
If we can have a reliable way to query for the historical Consumer validator set, that would be a good option to storing the historical validator set changes, which will take significant storage, and cannot or is not easy to prune.

[...]

PR #113, opened 2023-08-29
Slashing design considerations

### LINKED ISSUE #84: Add better documentation on slashing design
Not just how to detect, but what to do when it happens

See also #11 and #32.

maurolacy:
Part of #11. Maybe closes #84.

### DIFF
--- a/docs/ibc/Slashing.md
+++ b/docs/ibc/Slashing.md
@@ -7,7 +7,7 @@ proper slashing mechanics here.
 
 We are worried about a Byzantine consumer chain slashing arbitrary validators on the provider
 chain by writing false IBC packets. This could also be done via a bug in the State Machine
-and we in general don't want force the Provider to trust the Consumer state machine. However,
+and we in general don't want to force the Provider to trust the Consumer state machine. However,
 we should trust the Tendermint headers, which is the original source of double-signing
 evidence.
 
@@ -15,7 +15,7 @@ Rather than submit IBC packets for Slashing, we require submission of the duplic
 signatures from the Tendermint headers on the Consumer chain. These cannot be forged
 unless the private key is compromised.
 
-As of V1, the external staker must have a method to allow submitting such evidence of
+As of v1, the external staker must have a method to allow submitting such evidence of
 double-signing which can be verified and immediately slash all delegators of that
 validator.
 
@@ -66,3 +66,68 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
 (based on the unbonding period). Or just slash everyone bonded or unbonding, as the timestamps
 of the two votes may be wildly different, and they really shouldn't have trusted this
 cheating validator in the first place.
+
+## Trust Assumptions
+
+For the v1 implementation, we will assume that the Consumer chain is not and does not turn Byzantine,
+and that the Tendermint headers are valid and can be trusted.
+
+So, this is concerned with a malicious validator on the Consumer chain, double-signing to slash associated delegators
+on the Provider chain.
+In principle, nothing prevents a malicious validator on the Consumer to **intentionally** double-sign. So, the only
+thing we can do is limit the damage it can do.
+The amount of slashing for misbehaviour is defined by the slashing ratio. Additionally, we can rate limit the amount
+of slashing events, to one slashing event per day or similar. That will limit the damage a malicious validator
+can do to Provider funds, and will give time to cross-delegators on the Provider to unbond from that malicious validator.
+
+Another possibility is, a malicious validator on the Consumer double signing for profit, and trying to **avoid** being slashed.
+He could, by example, allocate all or most of his funds through cross-delegators on the Provider, and then tamper with
+the validator set updates, so that his public key, or associated block height and times, are invalid. This would prevent the
+Provider from slashing him, as the **provided evidence for misbehaviour would fail to verify**.
+This is a complex scenario, included here for completeness. It shows that the trust assumptions extend beyond the
+misbehaviour's evidence, and should include the validator set updates as well. Along with the Consumer chain itself.
+
+This indicates that it makes sense to re-utilize the same infrastructure and mechanisms that are used for
+communication between the Provider and the Consumer, for the specific case of slashing evidence submission.
+Both, slashing evidence handling and submission, and validator set updates, share similar trust assumptions and concerns.
+They should then be part of the same security model.
+
+Conversely, if, for complexity control and auditability, we decide to keep slashing evidence handling and submission
+separate from the rest of the Mesh Security infrastructure (in their own smart contracts on Consumer and Provider, by example),
+then we should make sure that slashing evidence handling and submission is done by the same entity that is responsible for
+validator set updates. So that they can be audited together, and the same trust assumptions apply to both.
+
+## Slashing Evidence Handling
+
+For v1, and for simplicity reasons, we will implement a slashing evidence handling mechanism as part of the existing infrastructure,
+namely, the established IBC channel between the `converter` contract on the Consumer and the `external-staking` contract on the Provider.
+This will be implemented as a new IBC packet type.
+
+The actual evidence will be submitted by the blockchain from a hook on the Evidence module, through a privileged (`sudo`) message
+to the `virtual-staking` contract, which already has sudo privileges. It will then be routed to the `converter` contract through a specific message,
+to be delivered to the `external-staking` contract on the Provider chain over IBC.
+This is similar to the way validator set updates are currently being implemented.
+
+The `external-staking` contract will then route the evidence to the `vault` contract, which will verify it and slash the associated delegators
+if the evidence is valid.
+The `vault` contract has the necessary information to verify the evidence, namely, the updated validator set.
+It also has the mapping between the offending validator and the associated cross-delegators.
+
+## Natively Staked Funds
+
+Mesh Security allows for simultaneous native- and cross-staking of funds. This is a powerful feature, and one of the main
+appeals behind it. However, it introduces some complications. In the case of a cross-slashing event, by example,
+funds may not be available on the vault for some or all of the impacted cross-delegators. For the simple (and legit) reason
+that they are being natively staked on the Provider chain at the same time.
+
+The `vault` contract will need to be able to handle this scenario.
+In the case of a slashed cross-delegator that doesn't have enough liquidity on the vault, it will need to unbond the funds
+from the Provider chain first, and then burn them to execute the slashing.
+
+A kind of "immediate unbonding" mechanism (and associated permission) could be needed for this.
+Alternatively, the vault can unbond the funds from the Provider chain and wait for the unbonding period to expire in order to slash them.
+This will be effectively the same, as during unbonding those funds are both, blocked from withdrawal, and not providing rewards.
+
+Finally, another option would be, for the vault to delegate the slashing in these cases to the blockchain itself.
+That is, by using and interacting with the Provider chain's Staking / Slashing module.
+This may be complicated to implement, as the slashing evidence and origin are effectively from another chain.
--- a/docs/ibc/Slashing.md
+++ b/docs/ibc/Slashing.md
@@ -74,8 +74,10 @@ and that the Tendermint headers are valid and can be trusted.
 
 So, this is concerned with a malicious validator on the Consumer chain, double-signing to slash associated delegators
 on the Provider chain.
+
 In principle, nothing prevents a malicious validator on the Consumer to **intentionally** double-sign. So, the only
 thing we can do is limit the damage it can do.
+
 The amount of slashing for misbehaviour is defined by the slashing ratio. Additionally, we can rate limit the amount
 of slashing events, to one slashing event per day or similar. That will limit the damage a malicious validator
 can do to Provider funds, and will give time to cross-delegators on the Provider to unbond from that malicious validator.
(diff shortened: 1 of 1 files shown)


docs/ibc/Slashing.md
```diff
@@ -66,3 +66,68 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+
+The actual evidence will be submitted by the blockchain from a hook on the Evidence module, through a privileged (`sudo`) message
+to the `virtual-staking` contract, which already has sudo privileges. It will then be routed to the `converter` contract through a specific message,
+to be delivered to the `external-staking` contract on the Provider chain over IBC.
+This is similar to the way validator set updates are currently being implemented.
+
+The `external-staking` contract will then route the evidence to the `vault` contract, which will verify it and slash the associated delegators
+if the evidence is valid.
+The `vault` contract has the necessary information to verify the evidence, namely, the updated validator set.
+It also has the mapping between the offending validator and the associated cross-delegators.
```
JakeHartnell, line 114:
```suggestion

The `vault` contract has the necessary information to verify the evidence, namely, the updated validator set.
It also has the mapping between the offending validator and the associated cross-delegators.
```

May also be good to comment on how this propagates across the Mesh. For example, if I'm cross-staking 100 Juno across five chains, and 5 Juno are slashed, then I'm now cross-staking 95 Juno. It needs to update other consumers / providers.

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,68 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+For the v1 implementation, we will assume that the Consumer chain is not and does not turn Byzantine,
+and that the Tendermint headers are valid and can be trusted.
+
+So, this is concerned with a malicious validator on the Consumer chain, double-signing to slash associated delegators
+on the Provider chain.
+In principle, nothing prevents a malicious validator on the Consumer to **intentionally** double-sign. So, the only
+thing we can do is limit the damage it can do.
+The amount of slashing for misbehaviour is defined by the slashing ratio. Additionally, we can rate limit the amount
+of slashing events, to one slashing event per day or similar. That will limit the damage a malicious validator
+can do to Provider funds, and will give time to cross-delegators on the Provider to unbond from that malicious validator.
```
JakeHartnell, line 81:
```suggestion
So, this is concerned with a malicious validator on the Consumer chain, double-signing to slash associated delegators
on the Provider chain.

In principle, nothing prevents a malicious validator on the Consumer to **intentionally** double-sign. So, the only
thing we can do is limit the damage it can do.

The amount of slashing for misbehaviour is defined by the slashing ratio. Additionally, we can rate limit the amount
of slashing events, to one slashing event per day or similar. That will limit the damage a malicious validator
can do to Provider funds, and will give time to cross-delegators on the Provider to unbond from that malicious validator.
```

JakeHartnell: Nit, I like spaces between paragraphs because I usually read these in me text editor.

docs/ibc/Slashing.md
```diff
@@ -7,15 +7,15 @@ proper slashing mechanics here.
 
 We are worried about a Byzantine consumer chain slashing arbitrary validators on the provider
 chain by writing false IBC packets. This could also be done via a bug in the State Machine
-and we in general don't want force the Provider to trust the Consumer state machine. However,
+and we in general don't want to force the Provider to trust the Consumer state machine. However,
 we should trust the Tendermint headers, which is the original source of double-signing
 evidence.
 
 Rather than submit IBC packets for Slashing, we require submission of the duplicate
 signatures from the Tendermint headers on the Consumer chain. These cannot be forged
```
jtremback, line 15: Don't you want these submitted and verified on the provider?

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,68 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
 (based on the unbonding period). Or just slash everyone bonded or unbonding, as the timestamps
 of the two votes may be wildly different, and they really shouldn't have trusted this
 cheating validator in the first place.
+
+## Trust Assumptions
+
+For the v1 implementation, we will assume that the Consumer chain is not and does not turn Byzantine,
```
jtremback, line 72: The exact trust assumptions are:

- We cannot trust the consumer chain's application, since this could be malicious.
- We are forced to trust the consumer chain's Comet or Tendermint process, since they could just double sign anyway if they are malicious.

This has some interesting implications. Validators can run any consumer chain application they want without worry, but they must be very careful if the application requires a fork of Comet.

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,68 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+and that the Tendermint headers are valid and can be trusted.
+
+So, this is concerned with a malicious validator on the Consumer chain, double-signing to slash associated delegators
+on the Provider chain.
+In principle, nothing prevents a malicious validator on the Consumer to **intentionally** double-sign. So, the only
+thing we can do is limit the damage it can do.
+The amount of slashing for misbehaviour is defined by the slashing ratio. Additionally, we can rate limit the amount
+of slashing events, to one slashing event per day or similar. That will limit the damage a malicious validator
```
jtremback, line 80: This is a very bad idea IMO. If I wanted to attack a consumer chain without being slashed and you had this policy in place, I would simply create a bunch of different validators, stake to them, do the attack, and then only be slashed once.

I think a slash needs to be a slash, especially if we are cryptographically verifying it. If a user delegates to a malicious validator and then gets slashed, this is their own fault. This is why they are getting staking rewards

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,68 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+can do to Provider funds, and will give time to cross-delegators on the Provider to unbond from that malicious validator.
+
+Another possibility is, a malicious validator on the Consumer double signing for profit, and trying to **avoid** being slashed.
+He could, by example, allocate all or most of his funds through cross-delegators on the Provider, and then tamper with
+the validator set updates, so that his public key, or associated block height and times, are invalid. This would prevent the
```
jtremback, line 85: In ICS, we are probably not going to trust anything from the consumer chain application, for simplicity of security analysis.

What this means is that the baseline slashing policy is this: While you are delegated to, undelegating from, or redelegating from a validator, you will be slashed if valid evidence of double signing by that validator is submitted (even if it is from a long time ago). This is somewhat stricter than Cosmos staking currently is, but we think it is OK.

We may be able to relax this somewhat, for example by excluding very old evidence, but these are relaxations of the baseline slashing policy above, and may not work in every case. We can stay in touch and let you know what we do.

docs/ibc/Slashing.md
JakeHartnell: Agree with this. If you don't like it, use a different slashing contract. :shrug:

docs/ibc/Slashing.md
JakeHartnell: Personally, I think this acceptable for the MVP slashing contract implementation.

docs/ibc/Slashing.md
JakeHartnell: Yes, I think we do.

https://github.com/osmosis-labs/mesh-security/pull/113#issuecomment-1697854435

maurolacy: Yes, the proposal is to submit the evidence of misbehaviour to the Provider, for verification there.

docs/ibc/Slashing.md
maurolacy: I think you are right. Will modify.

docs/ibc/Slashing.md
maurolacy: Yes, this is something we need to address. Will think about it and document it.

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,71 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+
+The actual evidence will be submitted by the blockchain from a hook on the Evidence module, through a privileged (`sudo`) message
+to the `virtual-staking` contract, which already has sudo privileges. It will then be routed to the `converter` contract through a specific message,
+to be delivered to the `external-staking` contract on the Provider chain over IBC.
+This is similar to the way validator set updates are currently being implemented.
+
+The `external-staking` contract will then route the evidence to the `vault` contract, which will verify it and slash the associated delegators
+if the evidence is valid.
+
+The `vault` contract has the necessary information to verify the evidence, namely, the updated validator set.
```
alpe, line 116: Are we prepared to have historic versions of the valset? I know that valset updates are not propagated from the consumer chain, currently

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,71 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+In the case of a slashed cross-delegator that doesn't have enough liquidity on the vault, it will need to unbond the funds
+from the Provider chain first, and then burn them to execute the slashing.
+
+A kind of "immediate unbonding" mechanism (and associated permission) could be needed for this.
+Alternatively, the vault can unbond the funds from the Provider chain and wait for the unbonding period to expire in order to slash them.
+This will be effectively the same, as during unbonding those funds are both, blocked from withdrawal, and not providing rewards.
+
+Finally, another option would be, for the vault to delegate the slashing in these cases to the blockchain itself.
+That is, by using and interacting with the Provider chain's Staking / Slashing module.
+This may be complicated to implement, as the slashing evidence and origin are effectively from another chain.
```
alpe, line 136: There is also the scenario to consider where slashing happens on the provider side and affects the staked collateral. A validator may become tombstoned, too.

docs/ibc/Slashing.md
```diff
@@ -66,3 +66,71 @@ evidence. Or we just accept any age and just use the age of the evidence to deci
...
+
+This indicates that it makes sense to re-utilize the same infrastructure and mechanisms that are used for
+communication between the Provider and the Consumer, for the specific case of slashing evidence submission.
+Both, slashing evidence handling and submission, and validator set updates, share similar trust assumptions and concerns.
+They should then be part of the same security model.
+
+Conversely, if, for complexity control and auditability, we decide to keep slashing evidence handling and submission
+separate from the rest of the Mesh Security infrastructure (in their own smart contracts on Consumer and Provider, by example),
+then we should make sure that slashing evidence handling and submission is done by the same entity that is responsible for
+validator set updates. So that they can be audited together, and the same trust assumptions apply to both.
```
alpe, line 100: Do we need to consider the async nature of the IBC channel?
An edge case but what if a valset update is received after the evidence for a double signing of a new validator?

docs/ibc/Slashing.md
maurolacy: Valset updates are already being propagated. Will create a couple of issues to:

- Change updates to support validator status (active, inactive, tombstoned). May be useful for cross-unbonding, so that we know if we need to apply the unbonding period or not. 
- Implement historical valset data on the Provider. So that we can check validator pubkey at the misbehaviour height.

docs/ibc/Slashing.md
maurolacy: Yes. Will think about "Slashing propagation" next.

Tombstoning can be supported through validator updates. We can block cross-bonding to tombstoned validators. And allow for immediate unbonding.

docs/ibc/Slashing.md
maurolacy: Good point. If we don't have conclusive data for a validator at the offending height, what we can do is store the slashing as pending. And process it after the timeout period for validator updates has passed.

maurolacy:
> > One high level note that @sunnya97 mentioned to me priviously is that slashing should ideally be implemented as separate contract on the provider side that handles the evidence. This allows for people to build custom slashers in the future (for example, a slasher based on Celestia Data Availability, or some other custom functionality). I think this important to keep as a design goal as it makes Mesh more flexible.
> > Different providers may choose to use different slashers, but we should definitely ship with a default slasher.
> 
> This is a good point. Also, in terms of auditing and complexity control, it makes sense to separate the slashing impl, on the Consumer at least, from the rest of the mesh security impl.
> 
> Problem with supporting a different contract for slashing initially is, it will make the setup more complex: a different contract will require another IBC channel setup, and a counterparty contract on the Provider just to handle the communication.
> 
> Perhaps we can go with a single contract impl initially, but make the mesh-security-sdk ready to support a different contract address for submitting misbehaviour evidence and validator set updates. It'll then become a setup / configuration issue, that we can address later. Will think about it.

We have decided to go with a single contract impl first, for simplicity. The code can be modular, both in mesh-security-sdk and in the contracts, so that it can be refactored later into a separate contract for slashing evidence handling.

docs/ibc/Slashing.md
maurolacy: You raise a very important point: The Provider side blockchain must inform the vault contract or the local staking contract of slashing events (and possibly, validator ser updates) as well.

From the point of view of slashing, locally- and cross-staked assets must be considered and processed almost similarly.
Local slashing events don't need to be verified. But, the effects of local slashing events on collateral must be updated in the vault. And slashing propagation would need to be carried on for both, local- and cross-slashing events.

maurolacy:
Added a couple of sections on native chain slashing and slashing propagation.

Will refine it and extend it. The case of jailing (leaving the active validator set) and tombstoning (permanent jailing) need to be considered and documented as well.

PR #117, opened 2023-08-29
feat: Move ibc workaround methods in ExternalStaking to interface

jawoznia:
Example usage of interface to keep test workaround methods out of contract


contracts/provider/external-staking/src/contract.rs
```diff
@@ -55,6 +54,7 @@ pub struct ExternalStakingContract<'a> {
 #[contract]
 #[error(ContractError)]
 #[messages(cross_staking_api as CrossStakingApi)]
+#[messages(crate::test_methods as TestMethods)]
```
maurolacy, line 57: 👍🏼

maurolacy (approved):
LGTM. Would like to avoid the need for this, but good for clarity / clean code.

PR #120, opened 2023-09-05
bug: add missing import for macro

Anmol1696:
## Overview
When running `cargo wasm`, got error of missing import. Seems to be true in the CI as well. 
Add import for macro.


JakeHartnell (approved):
Thanks!

PR #122, opened 2023-09-12
F/validator historical query

### LINKED ISSUE #119: Implement validator set historical queries
Implement historical valset data on the Provider. So that we can check for validator pubkey at the misbehaviour height.

This can be done through an `SnapshotMap` or `IndexedShapshotMap`.
Alternatively, it can be done through a conventional `Map` or `IndexedMap`, with composite `(validator, height)` keys, for flexibility and maintenance (pruning of older keys).

**Note**: This may be difficult to do with the current impl, as the valset updates use CRDT for consistency over unordered channels.
Let's consider changing this into a `SnapshotMap` or composite key `Map`, relaxing consistency and order guarantees in the pursuit of functionality: historical, queryable data.

maurolacy:
Closes #119.

Turns out only this is required to have historical validator queries (if we don't mind losing history when validator is tombstoned).

This can be improved by doing a [lower bound](https://cplusplus.com/reference/algorithm/lower_bound/) binary search for a given height. Given that queries will be mostly from recent heights, and that the list will be generally short, and can even be pruned relatively easily, I don't think it's worth it. In fact, linear search will probably be more efficient in practice.

### DIFF
--- a/contracts/provider/external-staking/src/crdt.rs
+++ b/contracts/provider/external-staking/src/crdt.rs
@@ -271,7 +271,13 @@ mod tests {
         crdt.add_validator(
             &mut storage,
             "alice",
-            mock_update_pubkey("alice_pubkey_2", 201),
+            mock_update_pubkey("alice_pubkey_2", 202),
+        )
+        .unwrap();
+        crdt.add_validator(
+            &mut storage,
+            "alice",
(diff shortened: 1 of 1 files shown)


JakeHartnell (approved):
LGTM! Thanks!

PR #125, opened 2023-09-14
Batch rewards distribution

### LINKED ISSUE #77: Use batch distribute_rewards for cross-chain distribution efficiency
The converter contract has distribute_reward, which we use now, which is called once per each validator getting rewards and then issues an ibc packet. This is N contract calls and N ibc packets every epoch.

There is also a distribute_rewards method that takes a batch of validators (all N?). This means only one contract call (saving N-1 * 60k gas). We could also use this to make one ibc packet with all distribution info.

This requires much more complex withdraw logic in the virtual-staking contract, and first we need to be able to cover that with unit tests. This in turn is blocked on some sylvia updates, and likely some cw-multi-test updates.

But when we have those dependencies ready, this would be a nice optimization to tackle.

uint:
Closes #77 

There's this comment in the ticket:
> This requires much more complex withdraw logic in the virtual-staking contract, and first we need to be able to cover that with unit tests.

Given that the ticket talks about batching distribution and not withdrawing, I don't understand the need for changing withdraw logic. I feel like maybe there's an implication withdrawing could be optimized too, but I have no way of knowing for sure what was meant. If I'm missing something or there's further work to be done, please let me know.

### DIFF
--- a/contracts/consumer/converter/src/contract.rs
+++ b/contracts/consumer/converter/src/contract.rs
@@ -326,7 +326,7 @@ impl ConverterApi for ConverterContract<'_> {
             }))
             .add_message(make_ibc_packet(
                 &mut ctx,
-                ConsumerPacket::DistributeRewards {
+                ConsumerPacket::DistributeBatch {
                     rewards: payments,
                     denom,
                 },
--- a/contracts/consumer/converter/src/contract.rs
+++ b/contracts/consumer/converter/src/contract.rs
@@ -1,6 +1,6 @@
 use cosmwasm_std::{
     ensure_eq, to_binary, Addr, BankMsg, Coin, CosmosMsg, Decimal, Deps, DepsMut, Event, IbcMsg,
-    Reply, Response, SubMsg, SubMsgResponse, Validator, WasmMsg,
+    Reply, Response, SubMsg, SubMsgResponse, Uint128, Validator, WasmMsg,
 };
 use cw2::set_contract_version;
 use cw_storage_plus::Item;
@@ -290,15 +290,7 @@ impl ConverterApi for ConverterContract<'_> {
             .add_attribute("validator", &validator)
             .add_attribute("amount", rewards.amount.to_string());
 
-        // Create a packet for the provider, informing of distribution
-        let packet = ConsumerPacket::Distribute { validator, rewards };
-        let channel = IBC_CHANNEL.load(ctx.deps.storage)?;
-        let msg = IbcMsg::SendPacket {
-            channel_id: channel.endpoint.channel_id,
-            data: to_binary(&packet)?,
-            timeout: packet_timeout_rewards(&ctx.env),
-        };
-
+        let msg = make_ibc_packet(&mut ctx, ConsumerPacket::Distribute { validator, rewards })?;
         Ok(Response::new().add_message(msg).add_event(event))
     }
 
@@ -313,19 +305,32 @@ impl ConverterApi for ConverterContract<'_> {
         mut ctx: ExecCtx,
         payments: Vec<RewardInfo>,
     ) -> Result<Response, Self::Error> {
-        // TODO: Optimize this, when we actually get such calls
-        let mut resp = Response::new();
-        for RewardInfo { validator, reward } in payments {
-            let mut sub_ctx = ctx.branch();
-            sub_ctx.info.funds[0].amount = reward;
-            let r = self.distribute_reward(sub_ctx, validator)?;
-            // put all values on the parent one
-            resp = resp
-                .add_submessages(r.messages)
-                .add_attributes(r.attributes)
-                .add_events(r.events);
+        let config = self.config.load(ctx.deps.storage)?;
+        let denom = config.local_denom;
+
+        let summed_rewards: Uint128 = payments.iter().map(|reward_info| reward_info.reward).sum();
+        let sent = must_pay(&ctx.info, &denom)?;
+
+        if summed_rewards != sent {
+            return Err(ContractError::DistributeRewardsInvalidAmount {
+                sum: summed_rewards,
+                sent,
+            });
         }
-        Ok(resp)
+
+        Ok(Response::new()
+            .add_events(payments.iter().map(|reward_info| {
+                Event::new("distribute_reward")
+                    .add_attribute("validator", &reward_info.validator)
+                    .add_attribute("amount", reward_info.reward)
+            }))
+            .add_message(make_ibc_packet(
+                &mut ctx,
+                ConsumerPacket::DistributeRewards {
+                    rewards: payments,
+                    denom,
+                },
+            )?))
     }
 
     /// Valset updates.
@@ -374,4 +379,13 @@ impl ConverterApi for ConverterContract<'_> {
 
         Ok(resp)
     }
+}
+
+fn make_ibc_packet(ctx: &mut ExecCtx, packet: ConsumerPacket) -> Result<IbcMsg, ContractError> {
+    let channel = IBC_CHANNEL.load(ctx.deps.storage)?;
+    Ok(IbcMsg::SendPacket {
+        channel_id: channel.endpoint.channel_id,
+        data: to_binary(&packet)?,
+        timeout: packet_timeout_rewards(&ctx.env),
+    })
 }
--- a/contracts/consumer/converter/src/contract.rs
+++ b/contracts/consumer/converter/src/contract.rs
@@ -1,6 +1,6 @@
 use cosmwasm_std::{
-    ensure_eq, to_binary, Addr, BankMsg, Coin, CosmosMsg, Decimal, Deps, DepsMut, Event, Reply,
-    Response, SubMsg, SubMsgResponse, Uint128, Validator, WasmMsg,
+    ensure_eq, to_binary, Addr, BankMsg, Coin, CosmosMsg, Decimal, Deps, DepsMut, Event,
+    MessageInfo, Reply, Response, SubMsg, SubMsgResponse, Uint128, Validator, WasmMsg,
 };
 use cw2::set_contract_version;
 use cw_storage_plus::Item;
@@ -265,13 +265,9 @@ impl ConverterContract<'_> {
         Ok(msg.into())
     }
 
-    fn ensure_authorized(&self, ctx: &ExecCtx) -> Result<(), ContractError> {
-        let virtual_stake = self.virtual_stake.load(ctx.deps.storage)?;
-        ensure_eq!(
-            ctx.info.sender,
-            virtual_stake,
-            ContractError::Unauthorized {}
-        );
+    fn ensure_authorized(&self, deps: &DepsMut, info: &MessageInfo) -> Result<(), ContractError> {
+        let virtual_stake = self.virtual_stake.load(deps.storage)?;
+        ensure_eq!(info.sender, virtual_stake, ContractError::Unauthorized {});
 
         Ok(())
     }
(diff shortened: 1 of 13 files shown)


uint:
This is only missing tests on provider site, but I'm going to first try and add better abstractions to provider tests.

uint:
@maurolacy I've noticed non-batch distribution returns no error when an invalid validator is provided. Is that intended? Do we want the same behavior for batch?

maurolacy:
> @maurolacy I've noticed non-batch distribution returns no error when an invalid validator is provided. Is that intended? Do we want the same behavior for batch?

That's OK, because that validators list comes from the blockchain. Take a look at the `virtual-staking` contract, the `handle_epoch` and `reply_rewards` methods gather the rewards automatically from the blockchain.

contracts/consumer/converter/src/contract.rs
```diff
@@ -375,3 +380,12 @@ impl ConverterApi for ConverterContract<'_> {
         Ok(resp)
     }
 }
+
+fn make_ibc_packet(ctx: &mut ExecCtx, packet: ConsumerPacket) -> Result<IbcMsg, ContractError> {
```
maurolacy, line 384: Nit: Put this into the `ibc` module?

contracts/consumer/converter/src/contract.rs
```diff
@@ -313,19 +305,32 @@ impl ConverterApi for ConverterContract<'_> {
...
+        Ok(Response::new()
+            .add_events(payments.iter().map(|reward_info| {
+                Event::new("distribute_reward")
+                    .add_attribute("validator", &reward_info.validator)
+                    .add_attribute("amount", reward_info.reward)
+            }))
+            .add_message(make_ibc_packet(
+                &mut ctx,
+                ConsumerPacket::DistributeBatch {
+                    rewards: payments,
```
maurolacy, line 330: 👍🏼

contracts/provider/external-staking/src/ibc.rs
```diff
@@ -153,10 +153,16 @@ pub fn ibc_packet_receive(
...
         ConsumerPacket::Distribute { validator, rewards } => {
             let contract = ExternalStakingContract::new();
-            let evt = contract.distribute_rewards(deps, validator, rewards)?;
+            let evt = contract.distribute_rewards(deps, &validator, rewards)?;
             let ack = ack_success(&DistributeAck {})?;
             IbcReceiveResponse::new().set_ack(ack).add_event(evt)
         }
+        ConsumerPacket::DistributeBatch { rewards, denom } => {
+            let contract = ExternalStakingContract::new();
+            let evts = contract.distribute_rewards_batch(deps, &denom, &rewards)?;
```
maurolacy, line 162: Nit: Reverse the params order for consistency.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -503,6 +512,32 @@ impl ExternalStakingContract<'_> {
...
+            denom,
+            config.rewards_denom,
+            ContractError::InvalidDenom(config.rewards_denom)
+        );
+
+        rewards
+            .iter()
+            .map(|reward_info| {
+                self.distribute_rewards_unchecked(
+                    &mut deps,
```
maurolacy, line 533: Nice.

maurolacy (commented):
Good start. Please adjust / remove tests, as invalid validators shouldn't be possible.

maurolacy:
> Given that the ticket talks about batching distribution and not withdrawing, I don't understand the need for changing withdraw logic. I feel like maybe there's an implication withdrawing could be optimized too, but I have no way of knowing for sure what was meant. If I'm missing something or there's further work to be done, please let me know.

What's missing is to send a `DistributeBatch` message from the virtual-staking contract to the converter, instead of a bunch of `Distribute` messages.

I'm realising now that this might be complex to do. Let's discuss details offline.

maurolacy:
> > @maurolacy I've noticed non-batch distribution returns no error when an invalid validator is provided. Is that intended? Do we want the same behavior for batch?
> 
> That's OK, because that validators list comes from the blockchain. Take a look at the `virtual-staking` contract, the `handle_epoch` and `reply_rewards` methods gather the rewards automatically from the blockchain.

What we can do to err on the safe side, is to add a sender check to the `distribute_reward` / `distribute_rewards` handlers, so that the only authorised caller is the `virtual-staking` contract.

That way we'll prevent malicious actors on the Consumer sending small amounts with invalid validators to distribute, to mess with our mesh security impl.

uint:
@maurolacy How do we test the code in `virtual-staking` currently?

maurolacy:
> @maurolacy How do we test the code in `virtual-staking` currently?

Good question. Short answer is, we don't have a good way to test some parts of it at this level at the moment. That's why it's mostly untested.
Feel free to write some unit tests if you think it's worth it / useful. But some parts, like the reply handler behaviour, are difficult to test in this setup.

By the way, now that Sylvia has support for custom queries, we can test the sudo entry point better (there's a FIXME in the code for that).

On the bright side, the rewards distribution code was tested as part of mesh-security integration on a devnet, and it works.

uint:
> > @maurolacy How do we test the code in `virtual-staking` currently?
> 
> Good question. Short answer is, we don't have a good way to test some parts of it at this level at the moment. That's why it's mostly untested. Feel free to write some unit tests if you think it's worth it / useful. But some parts, like the reply handler behaviour, are difficult to test in this setup.

I have ideas about how to test this, but they might require some refactoring and bloat this PR even more. I think I'd be more comfortable getting this merged and starting a new PR.

maurolacy:
Looks good. Is this ready for (final) review? Can you mark it as such? Thanks.

uint:
> Looks good. Is this ready for (final) review? Can you mark it as such? Thanks.

I was going to give it a once-over before doing that.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -201,23 +201,34 @@ impl VirtualStakingContract<'_> {
...
 
-        // Send them to the converter, assigned to proper validator
-        let msg = converter_api::ExecMsg::DistributeReward { validator: target };
-        let msg = WasmMsg::Execute {
-            contract_addr: cfg.converter.into_string(),
-            msg: to_binary(&msg)?,
-            funds: vec![reward],
-        };
-        let resp = Response::new().add_message(msg);
+        let all_rewards = VALIDATOR_REWARDS_BATCH.push(deps.storage, target, reward.amount)?;
```
maurolacy, line 204: Since we're not sending the intermediate messages, this amount will be cumulative, which is wrong.
We need to keep track of the previous amount, to discount it from the latest balance here.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -201,23 +201,34 @@ impl VirtualStakingContract<'_> {
             return Ok(Response::new());
```
maurolacy, line 201: If for whatever reason the last validator on the list doesn't have any rewards, the batch will never be sent.

We need to change this check to account for `finished`. Also, we need to check that the current balance minus the previous one is zero.

maurolacy (commented):
Reviewed the `virtual-staking` integration and found some issues. Will review the rest next.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -190,34 +190,53 @@ impl VirtualStakingContract<'_> {
         _reply: SubMsgResponse,
     ) -> Result<Response, ContractError> {
         // Find the validator to assign the rewards to
-        let target = pop_target(deps.branch())?;
+        let (target, finished) = pop_target(deps.branch())?;
 
         // Find all the tokens received here (consider it rewards to that validator)
         let cfg = self.config.load(deps.storage)?;
+        let total = VALIDATOR_REWARDS_BATCH.total(deps.storage)?;
```
maurolacy, line 197: Are we sure this is initialised to zero on creation?

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -190,34 +190,53 @@ impl VirtualStakingContract<'_> {
...
-            .query_balance(env.contract.address, cfg.denom)?;
-        if reward.amount.is_zero() {
-            return Ok(Response::new());
+            .query_balance(env.contract.address, &cfg.denom)?
+            .amount
+            - total;
+
+        let mut resp = Response::new();
+
+        if reward.is_zero() && (!finished || total.is_zero()) {
```
maurolacy, line 206: I find this logic confusing. Why not skip the specific message if the reward is zero and continue.

And then process the batch (if there are rewards) when `finished`.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -256,8 +275,54 @@ fn calculate_rebalance(
...
+        })
+    }
+
+    fn get(&self, store: &mut dyn Storage) -> StdResult<Vec<RewardInfo>> {
+        self.rewards.load(store)
+    }
+
+    fn wipe(&self, store: &mut dyn Storage) {
+        self.rewards.remove(store);
+    }
```
maurolacy, line 318: Shouldn't `wipe` reset `total` to zero as well?

maurolacy (commented):
Looks better, but I think that there are still a couple of issues with the main logic.

uint:
> I have ideas about how to test this, but they might require some refactoring and bloat this PR even more. I think I'd be more comfortable getting this merged and starting a new PR.

Since the tests didn't really require refactoring (just a bunch of new code), and the lack of them makes this PR development painful, I'm going to merge the tests bit here after all. Sorry about the LoC!

contracts/consumer/virtual-staking/src/contract.rs
uint: Yeah, good catch.

contracts/consumer/virtual-staking/src/contract.rs
uint: No. I fixed that in the tests PR, but forgot to backport here. I'll just merge these things so I can test the code properly.

contracts/consumer/virtual-staking/src/contract.rs
uint: I'll try to make it simpler.

contracts/consumer/virtual-staking/src/contract.rs
uint: Here's a win: [the test I added for this](https://github.com/osmosis-labs/mesh-security/pull/125/files#diff-55aab61e430421dc3035cc9d8bcf2263cbc08138462de265da9a4b0efe4c3038R467-R483) (and more than this) has only 16 LoC and doesn't require comments to be readable!

contracts/consumer/virtual-staking/src/contract.rs
uint: I ended up focusing even more on avoiding storage reads when they're unnecessary, but I'm hoping the code is a little less confusing too. No more convoluted boolean expressions.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -183,41 +185,81 @@ impl VirtualStakingContract<'_> {
...
 
-        // Send them to the converter, assigned to proper validator
-        let msg = converter_api::ExecMsg::DistributeReward { validator: target };
-        let msg = WasmMsg::Execute {
-            contract_addr: cfg.converter.into_string(),
-            msg: to_binary(&msg)?,
-            funds: vec![reward],
+        let new_reward_amount = total - BATCH.total(deps.storage)?;
+
+        let all_rewards = |storage| {
```
maurolacy, line 209: This logic is now more complex than before. 🙂

OK, I trust this is correct, based on the tests below.

Nice way of testing it, by the way.

maurolacy (approved):
LGTM. Nice tests!

PR #126, opened 2023-09-19
F/slashing

### LINKED ISSUE #11: Implement Slashing

maurolacy:
Closes #11. WIP. Publishing as draft for review / feedback.

TODO:

- [x] Cross slashing spec.
- [x] Cross slashing routing.
- [x] Validator tombstoning during slashing.
- [x] Slash users associated to validator.
- [x] Cross slashing propagation. Maintain mesh security invariants:
  - [x] Total slashable amounts.
  - [x] Liens.
- [x] Document slashing in detail. Scenarios / examples. (#127)
- [ ] Slashing related messages / handlers. (#128)
  - [ ] `burn`. Like `unstake` but without releasing funds. 
  - [ ] `slash_adjust`. For smart contracts accounting, but no blockchain calls (funds already slashed at blockchain level). (**wip**)
  - [x] slashing of the (unreleased) unbonding claims.
- [ ] Native slashing propagation. Similar to cross-slashing. (#129)
- [x] Scenario tests.
- [ ] Integration / multi-contract tests.

### DIFF
--- a/contracts/provider/external-staking/src/ibc.rs
+++ b/contracts/provider/external-staking/src/ibc.rs
@@ -9,7 +9,8 @@ use cosmwasm_std::{
 use cw_storage_plus::Item;
 use mesh_apis::ibc::{
     ack_success, validate_channel_order, AckWrapper, AddValidator, AddValidatorsAck,
-    ConsumerPacket, DistributeAck, ProtocolVersion, ProviderPacket, RemoveValidatorsAck,
+    ConsumerPacket, DistributeAck, ProtocolVersion, ProviderPacket, RemoveValidator,
+    RemoveValidatorsAck,
 };
 
 use crate::contract::ExternalStakingContract;
@@ -146,12 +147,18 @@ pub fn ibc_packet_receive(
         }
         ConsumerPacket::RemoveValidators(to_remove) => {
             let mut msgs = vec![];
-            for valoper in to_remove {
-                // Check that the validator is active (at height) and slash it if that is the case
-                // TODO: Needs Consumer chain's height
-                let active = contract
-                    .val_set
-                    .is_active_validator(deps.storage, &valoper)?;
+            for RemoveValidator {
+                valoper,
+                end_height,
+                end_time: _end_time,
+            } in to_remove
+            {
+                // Check that the validator is active at height and slash it if that is the case
+                let active = contract.val_set.is_active_validator_at_height(
+                    deps.storage,
+                    &valoper,
+                    end_height,
+                )?;
                 contract.val_set.remove_validator(deps.storage, &valoper)?;
                 if active {
                     // slash the validator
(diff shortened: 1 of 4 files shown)


maurolacy:
OK, that took a while, but I think we now have a correct slashing implementation, at the (vault's) accounting level. Both for slashing and slashing propagation.

Marking this as ready for review. Please take a look!

JakeHartnell (approved):
LGTM! Looking forward to this coming together. :partying_face:

maurolacy:
k, merging due to popular demand. ;-) I understand this is something that can be used for integration tests, etc.

Still, I would very much like a review here, even post-merge.

contracts/provider/external-staking/src/ibc.rs
```diff
@@ -145,11 +146,29 @@ pub fn ibc_packet_receive(
...
         ConsumerPacket::RemoveValidators(to_remove) => {
-            for valoper in to_remove {
+            let mut msgs = vec![];
+            for RemoveValidator {
+                valoper,
+                end_height,
+                end_time: _end_time,
+            } in to_remove
+            {
+                // Check that the validator is active at height and slash it if that is the case
```
uint, line 156: I'm not sure I understand this. It looks like `virtual-staking` informs `external-staking` that the validator has been **tombstoned**, but this triggers a slashing event for users staking on that validator?

uint (commented):
LGTM as far as I can tell, but I still don't understand this fully.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -658,15 +688,51 @@ impl ExternalStakingContract<'_> {
...
+            .stake
             .save(deps.storage, (&staker, &validator), &stake)?;
 
         Ok(())
     }
 
+    /// Slashes a validator.
+    ///
+    /// In test code, this is called from `test_handle_slashing`.
+    /// In non-test code, this is being called from `ibc_packet_receive` (in the `ConsumerPacket::RemoveValidators`
```
uint, line 707: I'm probably missing something, but should user slashing not get triggered when a validator itself is slashed?

contracts/provider/external-staking/src/ibc.rs
maurolacy: Yes, both tombstoning and jailing always imply slashing.

contracts/provider/external-staking/src/contract.rs
maurolacy: There's no specific slashing event. As mentioned, slashing is part of validator tombstone because of double signing, or validator jailing because of being offline.

Issue #123, opened 2023-09-13
Improve rewards withdrawal process


maurolacy:
The current rewards withdrawal process would break if a validator was slashed or fell out of the validator set.

See

https://github.com/osmosis-labs/mesh-security/blob/8e520dfdd9bc716c2723e2b349bef47baf8446cd/contracts/consumer/virtual-staking/src/contract.rs#L78-L97

And this comment

https://github.com/osmosis-labs/mesh-security/blob/8e520dfdd9bc716c2723e2b349bef47baf8446cd/contracts/consumer/virtual-staking/src/contract.rs#L138-L150

Fixing this requiring keeping track of the active validator set on the Consumer (i.e. on the `virtual-staking` contract).

This is similar to the way the valset is being kept up-to-date in Tgrade. You can take a look at [`tgrade-valset`](https://github.com/confio/poe-contracts/tree/main/contracts/tgrade-valset) for inspiration / ideas.

uint:
I think this was dealt with in #137. @maurolacy Do we close this, or is there more to do?

maurolacy:
There's more to do, but will be dealt with during full valset state implementation (#118). Closing.

PR #132, opened 2023-10-02
virtual-staking: Add tests for validator withdrawal and rebalance

uint:


uint:
@maurolacy Could you take a look if these are good enough tests for jail/tombstone/removal for #123? https://github.com/osmosis-labs/mesh-security/pull/132/files#diff-55aab61e430421dc3035cc9d8bcf2263cbc08138462de265da9a4b0efe4c3038R558-R619 They currently fail, but that's expected.

JakeHartnell (approved):
LGTM!

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -468,18 +468,163 @@ pub fn sudo(
...
+    fn empty_bond() {
+        let (mut deps, _bond_status) = mock_dependencies();
+
+        let contract = VirtualStakingContract::new();
+        contract.quick_inst(deps.as_mut());
+
+        contract.quick_bond(deps.as_mut(), "val1", 5);
+        contract
+            .hit_epoch(deps.as_mut())
+            .assert_no_bonding()
```
maurolacy, line 500: Is this because `max_cap` by default is zero? Please add a comment / rename the test.

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -468,18 +468,163 @@ pub fn sudo(
...
+        let contract = VirtualStakingContract::new();
+        contract.quick_inst(deps.as_mut());
+        let denom = contract.config.load(&deps.storage).unwrap().denom;
+
+        bond_status.update(5u128, 0u128);
+        contract.quick_bond(deps.as_mut(), "val1", 10);
+        contract.quick_bond(deps.as_mut(), "val2", 40);
+        contract
+            .hit_epoch(deps.as_mut())
+            .assert_bond(&[("val1", (1u128, &denom)), ("val2", (4u128, &denom))])
```
maurolacy, line 533: Is this because of rebalance, right? Better add a comment, and / or change the test name.

I would also add a test in which the `max_cap` is enough.

maurolacy (approved):
Sorry for the delay in reviewing. These tests are great, thanks for adding them.

contracts/consumer/virtual-staking/src/contract.rs
uint: Yes. Good catch - the max_cap = 0 should definitely be explicit.

contracts/consumer/virtual-staking/src/contract.rs
uint: What would be a clearer test name here? `bond_rebalance`?

`simple_bond` is a test case in which there's enough bonded.

uint: I added a test case that's like `simple_bond`, but with 2 validators.

maurolacy: `bond_rebalance` is OK.

Issue #95, opened 2023-07-19
Improve multi-tests


maurolacy:
Follow-up on the current state of testing.

TODO:

- [x] #146
- [x] Remove Mocks / stub contracts. Replace mocks by the real contracts, and adjust tests.
  - [x] `vault`: `cross_staking_mock`
  - [x] `vault`: `local_staking_mock`
  - [ ] [`converter`: `virtual_staking_mock`](https://github.com/osmosis-labs/mesh-security/blob/main/contracts/consumer/converter/src/multitest/virtual_staking_mock.rs)
  - [x] [`native_staking`: `local_staking_proxy`](https://github.com/osmosis-labs/mesh-security/blob/main/contracts/provider/native-staking/src/multitest/local_staking_proxy.rs)
- [x] Add missing test cases / improve the state of unit / integration testing. Consumer side contracts in particular lack a good of number of tests.
  - #131
  - #132
- [ ] Send IBC messages when possible. Perhaps by modifying cw-multi-test so that it doesn't fail on unknown IBC messages, but just log them and ignore them.
- [ ] Improve test (through `cfg_attr(test, msg(exec))`) code. See https://github.com/CosmWasm/sylvia/issues/206.

maurolacy:
Closing as done. Will create follow-up issues for the pending bullet points.

maurolacy:
Re-opening, as virtual-staking-mock replacement is pending.

PR #135, opened 2023-10-05
vault: remove native staking mock

uint:
~~WIP~~ (done). Closes part of #95.


contracts/provider/vault/src/multitest.rs
```diff
@@ -494,12 +535,8 @@ fn stake_local() {
...
     );
 
-    local_staking
-        .unstake(
-            vault.contract_addr.to_string(),
-            user.to_owned(),
-            coin(100, OSMO),
-        )
+    proxy
+        .unstake(user.to_owned(), coin(100, OSMO))
```
uint, line 539: @maurolacy I can't manage to simulate an unstake with the real native staking contract. It looks like the real thing has an unbonding period and something external (a native Cosmos SDK module maybe?) sends it money. Is that what happens? Do we have a way to do that in tests?

maurolacy: I think you may be hitting a current limitation in multi-test. Take a look at the `ProcessQueue` "sudo" message. There are some examples in the `native-staking-proxy` contract tests.

maurolacy: There's also a cw-mt issue to improve on this by the way. See https://github.com/CosmWasm/cw-multi-test/issues/53.

uint: Yeah, I think I figured it out. I stumbled into `ProcessQueue` too.

contracts/provider/vault/src/multitest.rs
```diff
@@ -28,17 +28,52 @@ const SLASHING_PERCENTAGE: u64 = 10;
...
+    let mut app = MtApp::new(|router, _api, storage| {
         for (&user, amount) in std::iter::zip(users, amounts) {
             router
                 .bank
                 .init_balance(storage, &Addr::unchecked(user), coins(*amount, OSMO))
                 .unwrap();
         }
     });
+
+    set_chain_native_denom(&mut app, OSMO);
```
maurolacy, line 40: 👍🏼 I would just put it in place here, but not a bad idea to encapsulate it.

maurolacy (approved):
LGTM. Thanks for refactoring it.

JakeHartnell (approved):
:fire:

PR #136, opened 2023-10-06
Fix validator withdrawal follow-up

maurolacy:
Process jailing and tombstoning events in `virtual-staking`, adjusting the bonded amounts so that they are consistent with their on-chain slashing.

#132 tests can now be extended to these scenarios.

### DIFF
--- a/contracts/consumer/virtual-staking/src/contract.rs
+++ b/contracts/consumer/virtual-staking/src/contract.rs
@@ -1,8 +1,8 @@
 use std::cmp::Ordering;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 
 use cosmwasm_std::{
-    coin, ensure_eq, entry_point, to_binary, Coin, CosmosMsg, CustomQuery, DepsMut,
+    coin, ensure_eq, entry_point, to_binary, Coin, CosmosMsg, CustomQuery, Decimal, DepsMut,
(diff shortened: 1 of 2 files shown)


uint (approved):
I'm guessing this works, but it confuses me a bit. @maurolacy Is the following reasoning correct?

When Alice gets jailed, there might be a bond request coming in **and** a valset update saying Alice has been jailed, so we want to be on the lookout for that. On future epochs there will be no bond requests for Alice while she's jailed, so we don't really need to keep track of that. Eventually, once she's unjailed, bond requests will start coming in again - and that's really how `virtual_staking` knows she's back in the valset.

maurolacy:
> I'm guessing this works, but it confuses me a bit. @maurolacy Is the following reasoning correct?
> 1. When Alice gets jailed, there might be a bond request coming in **and** a valset update saying Alice has been jailed, so we want to be on the lookout for that.
> 2. On future epochs there will be no bond requests for Alice while she's jailed, so we don't really need to keep track of that.
> 3. Eventually, once she's unjailed, bond requests will start coming in again - and that's really how `virtual_staking` knows she's back in the valset.

2. and 3. should be no issue IMO. Regarding 1., the impl tries to account for it, but there may be errors, or an incomplete solution. Testing should make this clear. Will do that next.

JakeHartnell (approved):
:partying_face:

PR #137, opened 2023-10-09
F/slash jailed validators

maurolacy:
This slashes jailed validators as well as tombstoned ones, for consistency with Consumer side accounting; both on-chain, and in-contract (`virtual-staking`) after #136.


JakeHartnell (approved):
LGTM, can't wait to test all this.

maurolacy:
> LGTM, can't wait to test all this.

Hi, will do a release once #136, #138 and related follow-ups are merged, which should be soon.

PR #144, opened 2023-10-11
vault: toggle local staking on/off

### LINKED ISSUE #140: Disable native staking
Native staking would require a mesh-security-sdk impl for the Provider chain; to deal with native slashing.

Until that's implemented, let's add a way to disable native staking altogether, so that a version of mesh-security with cross-slashing implemented can be released and used robustly in the mean time.

uint:
Closes #140

I don't know if this matters, but this breaks the Vault API (instantiation endpoint).


contracts/provider/vault/src/error.rs
```diff
@@ -43,4 +43,7 @@ pub enum ContractError {
 
     #[error("The tx {0} exists but comes from the wrong address: {1}")]
     WrongContractTx(u64, Addr),
+
+    #[error("No local staking configured for this vault")]
```
maurolacy, line 47:
```suggestion
    #[error("No local staking configured")]
```

maurolacy (approved):
LGTM. Since this breaks the Vault API (instantiation endpoint), let's not add this change as part of the 0.7.x releases.

uint:
I thought I fixed these. Let me just tackle the errors.

uint:
[I found this re coverage upload error.](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954)

Looks like we'd need to [add that token](https://docs.codecov.com/docs/codecov-uploader#per-repository-token).

PR #149, opened 2023-10-16
Introduce `VirtualStakingMock`

uint:
Part of #95


uint:
I'm giving up on removing the [`virtual_staking_mock`](https://github.com/osmosis-labs/mesh-security/blob/main/contracts/consumer/converter/src/multitest/virtual_staking_mock.rs) from `converter`. I can't really find a sensible way to do it without custom module support from Sylvia. I instead suggest merging the module mock for use in the future.

maurolacy (approved):
LGTM.

Issue #124, opened 2023-09-13
Improve cross-bond/unbond process


maurolacy:
Improve cross-bonding by:

- [x] Avoiding bonding to an inexistent validator. This can be easily checked before processing a cross-bond.
- [x] Avoiding bondings to a tombstoned validator. This can also be checked before processing a cross-bond.

Improve cross-unbonding by:

- [x] Automatically cross-unbonding from a tombstoned validator. This may require some coordination with the remote chain, as unbonded assets might not be released yet, at the time of processing a tombstoning.
- [x] Allow for immediate unbonding from 'unbonded' validators. These are validators that are not in the active set, but that are not jailed nor tombstoned (just unfunded). This would require keeping track and storing of validator state (#118).

uint:
> Avoiding bonding to an inexistent validator. This can be easily checked before processing a cross-bond.

Could be missing something, but [the cross-bonding improvements seem to already be there](https://github.com/osmosis-labs/mesh-security/blob/main/contracts/provider/external-staking/src/contract.rs#L1031-L1036). This seems to error on an attempt to stake when the validator state is anything other than active, so as long as the states are synced up, it should work?

maurolacy:
Yes, you're right.

uint:
> Automatically cross-unbonding from a tombstoned validator. This may require some coordination with the remote chain, as unbounded assets might not be released yet, at the time of processing a tombstoning.

The only way I figured out how to do this is to:
1. add `unbonding_period: u64` to `virtual-staking::Config`,
2. have `virtual-staking` keep track of tombstoned validators for who we want to cross-unbond delegators,
3. process those cross-unbonds (extra IBC message to `external-staking`?) in `handle_epoch`, after the unbonding period has passed.

This feels like complexity, storage and communication overhead for a rather minor feature. Wouldn't a better place for this sort of thing be in some off-chain service, assuming this is something Osmosis really wants? For other cases, unstaking is a 2-step process anyway (as far as contracts are concerned).

uint:
> Allow for immediate unbonding from 'unbonded' validators.

Is this possible in all scenarios when a validator is marked `unbonded`? I imagine a validator who fell out of the valset is `unbonded` immediately, but still has to wait for the release of their stake?

maurolacy:
> > Allow for immediate unbonding from 'unbonded' validators.
> 
[...]

PR #152, opened 2023-10-19
Immediately release stake when possible

uint:
Part of #124 (last bullet point)


contracts/provider/external-staking/src/contract.rs
```diff
@@ -370,9 +370,18 @@ impl ExternalStakingContract<'_> {
...
+        );
+
         // FIXME? Release period being computed after successful IBC tx
         // (Note: this is good for now, but can be revisited in v1 design)
-        let release_at = env.block.time.plus_seconds(config.unbonding_period);
+        let release_at = if immediate_release {
+            env.block.time
+        } else {
+            env.block.time.plus_seconds(config.unbonding_period)
+        };
```
maurolacy, line 384: 👍🏼

contracts/provider/external-staking/src/contract.rs
```diff
@@ -391,9 +400,18 @@ impl ExternalStakingContract<'_> {
         self.distribution
             .save(deps.storage, &tx_validator, &distribution)?;
 
+        let unbond_msg = if immediate_release {
+            // since the validator is unbonded, we shouldn't have to
+            // wait for the unbonding period to finish to unstake
+            self.withdraw_unbonded_inner(deps.branch(), env, &tx_user)?
+                .0
```
maurolacy, line 407: I don't think this is needed. Your approach above of changing the release time is enough IMO.

The funds will be there for immediate withdrawal, and (similarly than in the case of the release of retained funds) it's the user's responsibility to then claim them.

I would simplify this, even if the user's experience can suffer a bit, for simplicity and consistency.

contracts/provider/external-staking/src/crdt.rs
```diff
@@ -393,6 +394,14 @@ impl<'a> CrdtState<'a> {
         self.validators.save(storage, valoper, &validator_state)?;
         Ok(())
     }
+
+    pub fn validator_state(&self, storage: &dyn Storage, valoper: &str) -> StdResult<State> {
```
maurolacy, line 398: 👍🏼 API is a little confusing here, but I like this function. Perhaps 
 `validator_at_height` above can be refactored into `validator_state_at_height`, with the same return value.

maurolacy (approved):
Looks good. Please simplify it for consistency.

contracts/provider/external-staking/src/crdt.rs
uint: I can't do that since some callers of `validator_at_height` rely on data that's available in `ValState`, but not in `State`.

contracts/provider/external-staking/src/test_methods.rs
```diff
@@ -28,6 +28,25 @@ pub trait TestMethods {
         time: u64,
     ) -> Result<Response, Self::Error>;
 
+    /// Sets validator as `unbonded`.
+    #[msg(exec)]
+    fn test_remove_validator(
```
maurolacy, line 33: This is OK. Tried to add a generic `test_valset_update` but the macros clashed with the `too_many_arguments` warning.

maurolacy (approved):
LGTM.

PR #153, opened 2023-10-20
Virtual staking valset updates

### LINKED ISSUE #118: Valset update: validator state
As a follow-up of #109, let's change valset updates to support validator status (active, unbonded, jailed, tombstoned).

The list of states comes from https://hub.cosmos.network/main/validators/validator-faq.html#what-are-the-different-states-a-validator-can-be-in.

We can have:

- `active`: In validator set.
- `unbonded`: Not in validator set, because of not enough bonded assets.
- `jailed`: Not in validator set, because of misbehaviour.
- `tombstoned`: Permanently not in validator set, because of misbehaviour.

This may be used during cross-unbonding; so that we know if we need to apply the unbonding period or not.
It can also be checked during cross-bonding, to prevent bonding to tombstoned validators.

**Note**: This may be difficult to implement with the current impl, as the valset updates use CRDT for consistency over unordered channels.

maurolacy:
Now closes #118.

Keeps a (simplified) list of `inactive` validators, to filter them out from the rewards gathering process.

No state keeping besides `bonded` and `inactive`, for simplicity. Can be extended to proper valset state keeping if / when needed.

### DIFF
--- a/contracts/consumer/converter/src/contract.rs
+++ b/contracts/consumer/converter/src/contract.rs
@@ -363,6 +363,7 @@ impl ConverterApi for ConverterContract<'_> {
         let channel = IBC_CHANNEL.load(ctx.deps.storage)?;
 
         let mut event = Event::new("valset_update");
+        let mut is_empty = true;
 
         if !additions.is_empty() {
             event = event.add_attribute(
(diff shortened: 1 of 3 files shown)


JakeHartnell (approved):
LGTM!

PR #155, opened 2023-10-26
Osmosis price oracle

### LINKED ISSUE #99: Price oracle contract
Instead of [simple-price-feed](https://github.com/osmosis-labs/mesh-security/tree/main/contracts/consumer/simple-price-feed), implement a proper price oracle contract that gets its info from a DEX such as Osmosis. This may require querying of price information over IBC.

Support multiple crypto currencies / crypto currencies pairs, both from a local (Consumer side) DEX, or from a remote (Consumer side / other) DEX, over IBC.

Use [TWAP](https://en.wikipedia.org/wiki/Time-weighted_average_price) or similar to get a stable price estimation for a crypto currency pair.

uint:
Maybe closes #99

- [x] check for outdated price data in consumer (store timestamp)
- [x] review how/when subscriptions are saved/removed in provider (IBC handshake vs explicit local endpoint?)
- [x] for twap, verify how `sdk.Dec` gotten from [`arithmetic_twap_to_now`](https://github.com/osmosis-labs/osmosis-rust/blob/main/packages/osmosis-std/src/types/osmosis/twap/v1beta1.rs#L357C12-L357C34) is stringified
- [ ] tests


contracts/osmosis-price-provider/src/state.rs
```diff
@@ -0,0 +1,99 @@
...
+pub struct Subscriptions {
+    by_denom: Item<'static, HashMap<String, Subscription>>,
+    inactive: Item<'static, Vec<IbcChannel>>,
+}
+
+impl Subscriptions {
+    pub(crate) const fn new() -> Self {
+        Self {
+            by_denom: Item::new("subscriptions_by_denom"),
+            inactive: Item::new("subscriptions_inactive"),
```
maurolacy, line 23: Nit: Use a shorter namespaces?

contracts/osmosis-price-provider/src/ibc.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use crate::{contract::OsmosisPriceProvider, error::ContractError};
+
+const PROTOCOL_NAME: &str = "mesh-security-price-feed";
+/// This is the maximum version of the price feed protocol that we support
+const SUPPORTED_IBC_PROTOCOL_VERSION: &str = "0.1.0";
+/// This is the minimum version that we are compatible with
+const MIN_IBC_PROTOCOL_VERSION: &str = "0.1.0";
+
+const TIMEOUT: u64 = 60 * 60;
```
maurolacy, line 23: Pricing info is bound to be dynamic / can change fast. Reduce this to 10 minutes or so?

contracts/osmosis-price-provider/src/ibc.rs
```diff
@@ -0,0 +1,172 @@
...
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn ibc_packet_receive(
+    _deps: DepsMut,
+    _env: Env,
+    _msg: IbcPacketReceiveMsg,
+) -> Result<IbcReceiveResponse, ContractError> {
+    // this contract only sends out update packets over IBC - it's not meant to receive any
+    Err(ContractError::IbcPacketRecvDisallowed)
+}
```
maurolacy, line 118: Hmm, I was expecting this would work the other way around: The Consumer queries for pricing information.

I mean, to handle frequency of updates, etc. I think it makes sense for the Consumer to initiate the flow. It's also more aligned with the concept of a query for pricing info.

contracts/osmosis-price-provider/src/state.rs
```diff
@@ -0,0 +1,99 @@
...
+        Ok(())
+    }
+
+    pub(crate) fn bind_channel(
+        &self,
+        storage: &mut dyn Storage,
+        channel: IbcChannel,
+        denom: String,
+        pool_id: u64,
+    ) -> Result<(), ContractError> {
```
maurolacy, line 57: Why do you need / want one channel per denom? Wouldn't make sense to have a single IBC channel established, and then query over it, where "denom" (token pair, actually) is just a query parameter?

contracts/osmosis-price-provider/src/contract.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::ibc::make_ibc_packet;
+use crate::state::{Config, Subscription, Subscriptions};
+
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+const BASE_ASSET: &str = "OSMO";
+
+const EPOCH_IN_SECS: u64 = 120;
+const LAST_EPOCH: Item<'static, Timestamp> = Item::new("last_epoch");
```
maurolacy, line 24: Put these into the `OsmosisPriceProvider` struct?

contracts/osmosis-price-provider/src/contract.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    #[msg(instantiate)]
+    pub fn instantiate(
+        &self,
+        ctx: InstantiateCtx,
+        admin: String,
+    ) -> Result<Response, ContractError> {
+        nonpayable(&ctx.info)?;
```
maurolacy, line 48: Good point about non-payable instantiation handlers.

contracts/osmosis-price-provider/src/contract.rs
```diff
@@ -0,0 +1,112 @@
...
+use sylvia::{contract, schemars};
+
+use crate::error::ContractError;
+use crate::ibc::make_ibc_packet;
+use crate::state::{Config, Subscription, Subscriptions};
+
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+const BASE_ASSET: &str = "OSMO";
```
maurolacy, line 21: Good for an initial version, but IMO this should be part of a "token_pair" query parameter.

That would allow for querying different token pair prices, like "USDC/JUNO", etc.

contracts/osmosis-price-provider/src/contract.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    #[msg(exec)]
+    pub fn bind(
+        &self,
+        ctx: ExecCtx,
+        denom: String,
+        channel: IbcChannel,
+        pool_id: u64,
+    ) -> Result<Response, ContractError> {
```
maurolacy, line 68: I would get rid of this, and keep a single IBC channel that allows for multiple token pair queries.

maurolacy (commented):
Initial review of the price provider. Will take a look at the consumer next.

contracts/consumer/remote-price-feed/src/contract.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub const fn new() -> Self {
+        Self {
+            native_per_foreign: Item::new("price"),
+        }
+    }
+
+    /// Sets up the contract with an initial price.
+    /// If the owner is not set in the message, it defaults to info.sender.
+    #[msg(instantiate)]
+    pub fn instantiate(&self, ctx: InstantiateCtx) -> Result<Response, ContractError> {
```
maurolacy, line 33: It would be good to support a number of token pairs during instantiation of this contract.

Along with an update frequency.

That's why I think it's better if the logic is the other way around: this contract queries regularly the Osmosis contract for pricing info, for each of the token pairs it has been setup / instantiated with.

uint:
> Initial review of the price provider. Will take a look at the consumer next.

Hey @maurolacy! I appreciate the review, but there's no need to go so in-depth when this is still a draft, pending a cleanup. I feel like the energy is better spent taking a [high level view](https://github.com/osmosis-labs/mesh-security/issues/99#issuecomment-1787446052) and verifying if this is what we want. I'll address the comments on Monday. Till then!

maurolacy:
Thanks for the pointer. Put some comments there as well.

packages/bindings/src/sudo.rs
```diff
@@ -4,3 +4,8 @@ use cosmwasm_schema::cw_serde;
 pub enum SudoMsg {
     Rebalance {},
 }
+
+#[cw_serde]
+pub enum PriceFeedProviderSudoMsg {
+    EndBlock {},
+}
```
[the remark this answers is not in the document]
uint: That's more error-throwing code we have to support. I can do it if you prefer.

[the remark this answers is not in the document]
maurolacy: Was taking a look at this, and it merits a refactor. For one, `Rebalance` needs to be renamed, to `HandleEpoch` or similar, as the entry points does more than rebalancing at this point.

Let's put everything into `SudoMsg`, and use generic names like `BeginBlock` and `EndBlock`. The implementor can then choose which one to use.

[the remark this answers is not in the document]
maurolacy: Just took a look at the SDK, and in fact the end-blocker is being used for "rebalance"

https://github.com/osmosis-labs/mesh-security-sdk/blob/cb5b4724a890c39983ce56b2900e69ce16b32233/x/meshsecurity/abci.go#L20-L32

So, we just need one sudo msg type, and it can be called `EndBlock`.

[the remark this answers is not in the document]
maurolacy: Or perhaps, `HandleEpoch`, as the call is not on each block.

uint:
Alright, I won't stall more. I'll try to add some tests in a separate PR for what I can, although there's very little non-IBC logic here.

packages/bindings/src/sudo.rs
```diff
@@ -0,0 +1,6 @@
+use cosmwasm_schema::cw_serde;
+
+#[cw_serde]
+pub enum SudoMsg {
+    EndBlock {},
+}
```
maurolacy, line 6: Nit: Take a look at #161. According to that, this should now go into `packages/api/src/price_feed_api.rs`.

We can do that refactoring later.

contracts/consumer/remote-price-feed/README.md
```diff
@@ -0,0 +1,12 @@
...
+This implements the [price feed API](../../../packages/apis/src/price_feed_api.rs).
+
+This contract provides exchange rate data by contacting an [Osmosis Price Provider](../osmosis-price-provider).
+
+A single trading pair has to be configured on instantiation, along with the IBC endpoint. In case multiple trading pairs need to be synced, multiple contracts must be deployed.
+
+For this contract to work correctly:
+
+- An IBC connection to [Osmosis Price Provider](../osmosis-price-provider) must be opened.
+- A`SudoMsg::EndBlock {}` must be sent to the contract regularly. This will allow the contract to request regular updates of locally stored data.
```
maurolacy, line 12:
```suggestion
- A `SudoMsg::EndBlock {}` must be sent to the contract regularly. This will allow the contract to request regular updates of locally stored data.
```

contracts/osmosis-price-provider/src/error.rs
```diff
@@ -0,0 +1,31 @@
+use cosmwasm_std::StdError;
+use cw_utils::PaymentError;
+use mesh_apis::ibc::VersionError;
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ContractError {
```
maurolacy, line 7: Some of these errors are currently unused, and can be removed.

contracts/osmosis-price-provider/src/contract.rs
```diff
@@ -0,0 +1,67 @@
...
+        base: impl Into<String>,
+        quote: impl Into<String>,
+    ) -> Result<Decimal, ContractError> {
+        let querier = TwapQuerier::new(&deps.querier);
+
+        Decimal::from_str(
+            &querier
+                .arithmetic_twap_to_now(pool_id, base.into(), quote.into(), None)?
+                .arithmetic_twap,
+        )
```
maurolacy, line 64: Looks good. Let's confirm this with the Osmosis team at some point, and adjust if needed.

contracts/consumer/remote-price-feed/src/contract.rs
```diff
@@ -0,0 +1,126 @@
...
+        trading_pair: TradingPair,
+        auth_endpoint: AuthorizedEndpoint,
+        epoch_in_secs: u64,
+        price_info_ttl_in_secs: u64,
+    ) -> Result<Response, ContractError> {
+        nonpayable(&ctx.info)?;
+
+        set_contract_version(ctx.deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+        self.last_epoch
+            .save(ctx.deps.storage, &Timestamp::from_seconds(0))?;
```
maurolacy, line 57: Does this mean 'never'? You can as well save the current block time here.

contracts/consumer/remote-price-feed/src/contract.rs
```diff
@@ -0,0 +1,126 @@
...
+pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
+pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub struct RemotePriceFeedContract {
+    pub channel: Item<'static, IbcChannel>,
+    pub trading_pair: Item<'static, TradingPair>,
+    pub price_info: Item<'static, PriceInfo>,
+    pub last_epoch: Item<'static, Timestamp>,
+    pub epoch_in_secs: Item<'static, u64>,
+    pub price_info_ttl_in_secs: Item<'static, u64>,
```
maurolacy, line 25: Makes sense. Setting a TTL that is slightly larger than the epoch will help with handling delays in the price update process.

Perhaps all this can be derived from a single parameter. Including the IBC timeout as well. But, let's leave that for another iteration.

contracts/consumer/remote-price-feed/src/contract.rs
```diff
@@ -0,0 +1,126 @@
...
+
+#[contract]
+#[messages(price_feed_api as PriceFeedApi)]
+impl PriceFeedApi for RemotePriceFeedContract {
+    type Error = ContractError;
+
+    /// Return the price of the foreign token. That is, how many native tokens
+    /// are needed to buy one foreign token.
+    #[msg(query)]
+    fn price(&self, ctx: QueryCtx) -> Result<PriceResponse, Self::Error> {
```
maurolacy, line 77: Good implementation.

contracts/consumer/remote-price-feed/src/contract.rs
```diff
@@ -0,0 +1,126 @@
...
+            Err(ContractError::OutdatedPriceData)
+        }
+    }
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn sudo(deps: DepsMut, env: Env, msg: SudoMsg) -> Result<Response, ContractError> {
+    match msg {
+        SudoMsg::EndBlock {} => {
+            let contract = RemotePriceFeedContract::new();
```
maurolacy, line 98: It would be good to put all this into a helper function, for testing and modularisation.

contracts/consumer/remote-price-feed/src/contract.rs
```diff
@@ -0,0 +1,126 @@
...
+            let TradingPair {
+                pool_id,
+                base_asset,
+                quote_asset,
+            } = contract.trading_pair.load(deps.storage)?;
+            let channel = contract
+                .channel
+                .may_load(deps.storage)?
+                .ok_or(ContractError::IbcChannelNotOpen)?;
```
maurolacy, line 107: You can load this inside the if, to save some gas.

maurolacy (approved):
Pretty good. Looking forward to some tests, and we can merge.

packages/bindings/src/sudo.rs
maurolacy: Another thing we can do here is to name this enum `HandelEpoch`. To signal the fact that it doesn't need to be called at each block.

The SDK can call it once per minute or so, and all of the timeouts / update logic will still work.

contracts/consumer/remote-price-feed/src/contract.rs
uint: No, this is the time of the last update. `0` stands for year 1970 and is pretty much guaranteed to mean "out of date", so that an update is scheduled after instantiation. If we change it to "now", the next update will be at `now + epoch_in_secs`, so it might introduce an unnecessary lag if anything.

uint: TBH if instead of using "end block" we can configure an interval in the native Cosmos module, this whole "epoch length" dance becomes unnecessary here.

maurolacy: Yes, but it doesn't hurt either, and allows for different contract to have different update periods while using the same SDK.

It's also slightly more robust.

PR #160, opened 2023-11-08
Cw update

maurolacy:
Update CosmWasm and tools to latest compatible versions.


uint:
Belatedly, this probably didn't update anything. What you did was narrow [version requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) to give dependency resolution less to work with. The actual version numbers reside in `Cargo.lock` and are updated with `cargo update`.

maurolacy:
Since there are no Cargo.lock changes, you're exactly right.

This was an attempt to update everything to latest CosmWasm 1.5.0, which failed because of some outdated deps.

Anyway, I think specifying precise deps versions is better, so. Will belatedly edit the PR's title.

PR #162, opened 2023-11-11
Native slashing acct

### LINKED ISSUE #129: Native slashing accounting
In the case of a native slashing event, the `vault` contract needs to be notified to perform the proper slashing adjustment over the associated users / funds.

This is similar to the cross-slashing case, but requires a handler (sudo message) on the Provider side contracts (`native-staking` probably)

maurolacy:
Closes #129.

Proposed jailing API is tentative and can be changed / adjusted if needed.

Pre- vs. post-slashing behaviour of the delegated amounts (at the moment of the jailing handler call) must be established / defined based on on-chain checks, and / or analysis.

TODO:

- [x] Tests.


maurolacy:
Merging this, as it's already in good shape. We can always revisit the design / API later.

PR #164, opened 2023-11-13
support multiple slash ratios

### LINKED ISSUE #151: Support multiple slash ratios
In Cosmos based blockchains there are typically two slash ratios, one slashing for being offline, and another for double signing.

Let's add support for those on the Provider. This is already required by the current slashing during jailing impl (https://github.com/osmosis-labs/mesh-security/blob/331ace589c0b693de075368e5875777e9d15fcda/contracts/provider/external-staking/src/contract.rs#L499).

uint:
Closes #151 

The changes to `native-staking` are there to support the modified `LocalStakingApi::max_slash`.

### DIFF
--- a/contracts/provider/external-staking/src/contract.rs
+++ b/contracts/provider/external-staking/src/contract.rs
@@ -26,7 +26,7 @@ use crate::msg::{
     StakeInfo, StakesResponse, TxResponse, ValidatorPendingRewards,
 };
 use crate::stakes::Stakes;
-use crate::state::{Config, Distribution, Stake};
+use crate::state::{Config, Distribution, MaxSlashing, Stake};
 
 pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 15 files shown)


packages/apis/src/local_staking_api.rs
```diff
@@ -5,7 +5,8 @@ use sylvia::{interface, schemars};
 
 #[cw_serde]
 pub struct MaxSlashResponse {
-    pub max_slash: Decimal,
+    pub max_slash_dsign: Decimal,
```
maurolacy, line 8: Nit: We can get rid of the `max`, here and elsewhere; or replace it by `ratio`.

The response can be called `SlashRatiosResponse`. And the query can be `SlashRatios`.

We can do that in a follow-up.

contracts/provider/external-staking/src/contract.rs
```diff
@@ -472,7 +477,8 @@ impl ExternalStakingContract<'_> {
                 .tombstone_validator(deps.storage, valoper, height, time)?;
             if active {
                 // Slash the validator (if bonded)
-                let slash_msg = self.handle_slashing(&env, deps.storage, valoper)?;
+                let slash_msg =
+                    self.handle_slashing(&env, deps.storage, valoper, SlashingReason::DoubleSign)?;
```
maurolacy, line 481: 👍🏼

maurolacy (approved):
LGTM.

PR #179, opened 2023-12-05
Slash amounts support

maurolacy:
Related to https://github.com/osmosis-labs/mesh-security-sdk/pull/125.

Adds support for slashing based on historical evidence (double signing slashing), through slashing amounts instead of slash ratios.

There are still some limitations and there's room for improvement in the contracts side impls. Will be documented in a number of follow-up issues.

**TODO**:

- [x] Fix / adapt / extend tests.
- [ ] Improve coverage / Add tests.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -84,4 +84,4 @@ jobs:
         with:
           toolchain: 1.70.0
           command: clippy
-          args: --all-targets -- -D warnings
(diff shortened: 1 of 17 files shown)


JakeHartnell (approved):
:tada:

JakeHartnell (dismissed):
Needs test fixes but otherwise looks good. Sorry for the earlier approval. :joy:

PR #188, opened 2024-04-28
Sylvia upgrade

JakeHartnell:
Attempts to upgrade to Sylvia `v0.10`, following roughly what's in this article: https://medium.com/confio/sylvia-1-0-0-45094b6ad8ba

The rational for `v0.10` over `1.0` is that the Starship testing infra chains don't yet support CW `2.0` I believe.

Doesn't fully work yet... would appreciate some help.

Builds off of #187, but separate for easier review.


contracts/consumer/converter/src/contract.rs
```diff
@@ -272,6 +272,7 @@ impl ConverterContract<'_> {
             }
         );
 
+        // FIXME not sure how to get this to compile with latest sylvia
         // get the price value (usage is a bit clunky, need to use trait and cannot chain Remote::new() with .querier())
         // also see https://github.com/CosmWasm/sylvia/issues/181 to just store Remote in state
         use price_feed_api::Querier;
```
JakeHartnell, line 278: @hashedone not sure what's changed here, but this part is very broken. Struggling to find examples for this `Querier` / `Remote` stuff...

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -604,34 +454,165 @@ impl VirtualStakingApi for VirtualStakingContract<'_> {
...
+     * 3. If the sum <= max_cap then use collected requests as is
+     * 4. If the sum > max_cap,
+     *   a. calculate multiplier Decimal(max_cap / sum)
+     *   b. multiply every element of the collected requests in place.
+     * 5. Find diff between collected (normalized) requests and last bonding amounts (which go up, which down).
+     * 6. Transform diff into unbond and bond requests, sorting so all unbond happen first
+     */
+    fn handle_epoch(
+        &self,
+        ctx: SudoCtx<VirtualStakeCustomQuery>,
```
JakeHartnell, line 476: These two sudo methods (`handle_epoch` and `valset_update`) require custom queries and messages. Any guidance or suggestions on how to handle this would be greatly appreciated. 🙏 

Examples or commits would be 🍰 .

contracts/consumer/virtual-staking/src/multitest.rs
```diff
@@ -15,10 +15,9 @@ struct SetupArgs<'a> {
...
-    price_feed:
-        mesh_simple_price_feed::contract::multitest_utils::SimplePriceFeedContractProxy<'a, MtApp>,
-    converter: mesh_converter::contract::multitest_utils::ConverterContractProxy<'a, MtApp>,
-    virtual_staking: contract::multitest_utils::VirtualStakingContractProxy<'a, MtApp>,
+    price_feed: mesh_simple_price_feed::contract::sv::mt::SimplePriceFeedContractProxy<'a, MtApp>,
+    converter: mesh_converter::contract::sv::mt::ConverterContractProxy<'a, MtApp>,
+    virtual_staking: contract::sv::mt::VirtualStakingContractProxy<'a, MtApp>,
 }
 
 fn setup<'a>(app: &'a App<MtApp>, args: SetupArgs<'a>) -> SetupResponse<'a> {
```
JakeHartnell, line 23: I'm having a hard time with these setup functions... and not just in this file. It's generic and trait hell.

Getting tones of errors like this:
```
error[E0107]: trait takes 9 generic arguments but 1 generic argument was supplied
  --> contracts/provider/native-staking-proxy/src/multitest.rs:59:16
   |
59 | ) -> AnyResult<VaultContractProxy<'app, MtApp>> {
   |                ^^^^^^^^^^^^^^^^^^       ----- supplied 1 generic argument
   |                |
   |                expected 9 generic arguments
   |
note: trait defined here, with 9 generic parameters: `BankT`, `ApiT`, `StorageT`, `CustomT`, `WasmT`, `StakingT`, `DistrT`, `IbcT`, `GovT`
  --> /Users/meow/Dev/Mesh/mesh-security/contracts/provider/vault/src/contract.rs:69:6
   |
66 | #[contract]
   | -----------
...
69 | impl VaultContract<'_> {
   |      ^^^^^^^^^^^^^
help: add missing generic arguments
   |
59 | ) -> AnyResult<VaultContractProxy<'app, MtApp, ApiT, StorageT, CustomT, WasmT, StakingT, DistrT, IbcT, GovT>> {
   |                                              ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

error[E0782]: trait objects must include the `dyn` keyword
  --> contracts/provider/native-staking-proxy/src/multitest.rs:59:16
   |
59 | ) -> AnyResult<VaultContractProxy<'app, MtApp>> {
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
help: add `dyn` keyword before this trait
   |
59 | ) -> AnyResult<dyn VaultContractProxy<'app, MtApp>> {
   |                +++

```

Too tired to figure it out right now, but any help would be appreciated.

JakeHartnell: > expected 9 generic arguments

🥲

contracts/consumer/converter/src/contract.rs
jawoznia: Since `0.10.0` `Sylvia` generates all the code in the `sv` module. We decided not to populate the scope with all the stuff it generates and rather keep it in this module.

Currently both `interface` and `contract` generates the code like this.

```rust
pub mod sv {
    // messages
    // qoc code
    
    pub mod mt {
    }
}
```

So for your code you have to:

```diff
+ use price_feed_api::sv::Querier;
- use price_feed_api::Querier;
```

You can read about the changes in the release article https://medium.com/confio/sylvia-1-0-0-45094b6ad8ba

I noticed that the https://github.com/CosmWasm/sylvia/blob/main/MIGRATING.md is not fully updated with the changes. I will update it today. Sorry for inconvenience.

JakeHartnell: > I noticed that the https://github.com/CosmWasm/sylvia/blob/main/MIGRATING.md is not fully updated with the changes. I will update it today. Sorry for inconvenience.

This would be super helpful.

How do I derive the `Remote`? Read the issue here and not sure what to make of it: https://github.com/CosmWasm/sylvia/issues/181

jawoznia: Both `Remote` and `BoundQuerier` were moved to the `sylvia::types` module as they were always generated in the same way.

Here is example usage of `Remote`/`Querier`. If you find them insufficient and/or have some suggestions what is missing we will appreciate the feedback.
https://github.com/CosmWasm/sylvia/blob/main/sylvia/tests/querier.rs

contracts/consumer/virtual-staking/src/contract.rs
jawoznia: I checked the branch and found out that some `Ctx` types miss the `<VirtualStakeCustomQuery>` parameter and same for `CustomMsg` for `Response`.

From Rust POV `Response<Empty>` and `Response<VirtualStakeCustomMsg>` are different types and you have to make all the methods in the dispatch flow use the same types.

contracts/consumer/converter/src/contract.rs
jawoznia: Those missing `derives` were implemented in https://github.com/CosmWasm/sylvia/pull/194 and are still present.
https://docs.rs/sylvia/0.10.1/sylvia/types/struct.Remote.html

Which traits are missing for your use case?

contracts/consumer/virtual-staking/src/multitest.rs
jawoznia: `Proxy` type generated by `Sylvia` is generic over `cw_multi_test` modules while it would be much more user friendly to be generic over `cw_multi_test::App`.
I see the missing use case in our tests. I will inspect possibility to make it generic over the `cw_multi_test::App`.

jawoznia: For now to make it work you can provide the types used in the `cw_multi_test::App` that I see you are trying to use here
https://github.com/CosmWasm/cw-multi-test/blob/main/src/app.rs#L51

jawoznia: Even better. You are returning here the trait. Change it to https://docs.rs/sylvia/0.10.1/sylvia/multitest/struct.Proxy.html.

You will be able to provide the `MtApp` here as first parameter and your contract as a second generic parameter.

contracts/consumer/converter/src/contract.rs
```diff
@@ -30,8 +30,8 @@ pub struct ConverterContract<'a> {
 
 #[cfg_attr(not(feature = "library"), sylvia::entry_points)]
 #[contract]
-#[error(ContractError)]
-#[messages(converter_api as ConverterApi)]
+#[sv::error(ContractError)]
+#[sv::messages(converter_api as ConverterApi)]
```
jawoznia, line 34: Since `0.10.0` below is possible since we can deduce the interface name by converting the path to camel case

```suggestion
#[sv::messages(converter_api)]
```

contracts/consumer/virtual-staking/src/contract.rs
```diff
@@ -69,7 +71,7 @@ impl VirtualStakingContract<'_> {
     }
 
     /// The caller of the instantiation will be the converter contract
-    #[msg(instantiate)]
+    #[sv::msg(instantiate)]
     pub fn instantiate(&self, ctx: InstantiateCtx) -> Result<Response, ContractError> {
```
jawoznia, line 75: Since the contract is now using some custom message and query you have to use them in the methods signatures. Otherwise you get this errors with conversion from `DepsMut<'_, Empty>` to `DepsMut<'_, VirtualStakeCustomQuery>`

```suggestion
    pub fn instantiate(&self, ctx: InstantiateCtx<VirtualStakeCustomQuery>) -> Result<Response<VirtualStakeCustomMsg>, ContractError> {
```

contracts/consumer/converter/src/contract.rs
ethanfrey: Okay, please look at https://github.com/osmosis-labs/mesh-security/pull/188/commits/71dffdf517508341dfb89b5d1129842af6dc01c0 and let me know if that was the proper way to deal with it. I feel I am just forcing the type system rather then understanding anything

jawoznia (commented):
@JakeHartnell, @ethanfrey 

Please check 528c6e1bbf3ae7b2c206a619f8ad6365cfcd6baa with the update of `consumer/converter`. Tests are building and passing.

contracts/consumer/converter/src/multitest.rs
```diff
@@ -20,10 +27,9 @@ struct SetupArgs<'a> {
 }
 
 struct SetupResponse<'a> {
-    price_feed:
-        mesh_simple_price_feed::contract::multitest_utils::SimplePriceFeedContractProxy<'a, MtApp>,
-    converter: contract::multitest_utils::ConverterContractProxy<'a, MtApp>,
-    virtual_staking: virtual_staking_mock::multitest_utils::VirtualStakingMockProxy<'a, MtApp>,
+    price_feed: Proxy<'a, MtApp, SimplePriceFeedContract<'a>>,
```
ethanfrey, line 30: Okay, this is a lot of non-intuitive changes, but happy to use this as an example to fix others.

contracts/consumer/converter/src/multitest.rs
```diff
@@ -34,10 +40,9 @@ fn setup<'a>(app: &'a App<MtApp>, args: SetupArgs<'a>) -> SetupResponse<'a> {
         native_per_foreign,
     } = args;
 
-    let price_feed_code =
-        mesh_simple_price_feed::contract::multitest_utils::CodeId::store_code(app);
-    let virtual_staking_code = virtual_staking_mock::multitest_utils::CodeId::store_code(app);
-    let converter_code = contract::multitest_utils::CodeId::store_code(app);
+    let price_feed_code = PriceFeedCodeId::store_code(app);
```
ethanfrey, line 43: Usage actually looks simpler

contracts/consumer/converter/src/contract.rs
```diff
@@ -275,8 +274,11 @@ impl ConverterContract<'_> {
         // FIXME not sure how to get this to compile with latest sylvia
         // get the price value (usage is a bit clunky, need to use trait and cannot chain Remote::new() with .querier())
         // also see https://github.com/CosmWasm/sylvia/issues/181 to just store Remote in state
-        use price_feed_api::Querier;
-        let remote = price_feed_api::Remote::new(config.price_feed);
+        use price_feed_api::sv::Querier;
+        use sylvia::types::Remote;
+        // NOTE: Jan, this feels hacky... I'm not sure if this is the right way to do it
+        // Also, I am sticking in a random error type here, not what I will get (which is unknown)
+        let remote = Remote::<&dyn price_feed_api::PriceFeedApi<Error=StdError>>::new(config.price_feed);
```
ethanfrey, line 281: Is this line correct? I feel I am faking the Remote generic type so it compiles... is this the right way?

ethanfrey (commented):
Thanks for the help @jawoznia 

Left a few comments and one big question

contracts/consumer/converter/src/multitest.rs
jawoznia: This change is because we moved out of generating `Proxy` type per `contract` macro.

Instead we have a `sylvia::multitest::Proxy` type that is generic over `MtApp` on which the `SimplePriceFeedContract` is deployed. So the idea was to link via proxy some contract to the chain.

contracts/consumer/converter/src/contract.rs
jawoznia: This is correct, as `sylvia::types::Remote` being representation of some external Contract is generic over it.

Unfortunately Rust requires us to state the associated types defined on the `PriceFeedApi` trait which is not ideal, but we went with it as this way we were able to simplify support for generics and overall make the API less clunky (not without sacrifices as we can see).
In current state and in this usage you are using the `Remote` to instantiate `BoundQuerier` and query on contract.
It's fine to set `Error` as `StdError` as underneath it calls [`query_wasm_smart`](https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.QuerierWrapper.html#method.query_wasm_smart) which returns the `StdResult` so although it looks hacky it's all fine.

The problem however is with storing `Remote` as a field, which we didn't cover for the `Remote` generic over some type implementing the interface and we have to fix it.

ethanfrey:
@jawoznia I managed to fix a lot with your example. All code and tests compile properly - expect `vitrual-staking`. Some tests are failing (on asserts not equaling), but I will investigate the diff to try to figure out why.

The biggest issue is `virtual-staking` that uses custom query and message type. I made a number of changes to get consistency everywhere and it compiles and seems proper, except multitest fails. The biggest issue is when we use the app with custom msg/query (needed for virtual staking) and use it to instantiate another contract (that has no custom msg/query), then it fails like this:

```
error[E0308]: mismatched types
  --> contracts/consumer/virtual-staking/src/multitest.rs:42:79
   |
42 |     let converter_code = mesh_converter::contract::sv::mt::CodeId::store_code(app);
   |                          ---------------------------------------------------- ^^^ expected `&App<App<_, _, _, _, WasmKeeper<Empty, Empty>, _, _, _, _>>`, found `&App<App<BankKeeper, MockApi, MemoryStorage, ..., ...>>`
   |                          |
   |                          arguments to this function are incorrect
   |
   = note: expected reference `&sylvia::multitest::App<cw_multi_test::App<_, _, _, _, WasmKeeper<cosmwasm_std::Empty, cosmwasm_std::Empty>, _, _, _, _>>`
              found reference `&'a sylvia::multitest::App<cw_multi_test::App<BankKeeper, MockApi, MemoryStorage, FailingModule<mesh_bindings::VirtualStakeCustomMsg, mesh_bindings::VirtualStakeCustomQuery, cosmwasm_std::Empty>, WasmKeeper<mesh_bindings::VirtualStakeCustomMsg, mesh_bindings::VirtualStakeCustomQuery>>>`
```

contracts/provider/vault/src/multitest.rs
```diff
@@ -273,9 +273,16 @@ fn proxy_for_user<'a>(
...
 
 fn process_staking_unbondings(app: &App<MtApp>) {
-    let mut block_info = app.block_info();
-    block_info.time = block_info.time.plus_seconds(61);
-    app.set_block(block_info);
+    app.app_mut().update_block(|block| {
+        block.time = block.time.plus_seconds(61);
+        block.height += 13;
+    });
+    // This is deprecated as unneeded, but tests fail if it isn't here. What's up???
```
ethanfrey, line 280: @jawoznia another one as possible bug in the cosmwasm frameworks...

This is marked deprecated, saying I should just call update_block. But if I comment these lines out the tests fail.
I even tried to update the block with some huge diff (610.000.000 seconds) and it doesn't trigger the process queue.

Can you investigate? I know it is cw-multi-test, but still...

JakeHartnell (commented):
LGTM! Thanks @ethanfrey and @jawoznia for helping get this updated!

contracts/provider/vault/src/multitest.rs
jawoznia: If this is not yet an issue in `cw-multi-test` please consider making this an issue there. Darek should definitely check it.

PR #190, opened 2024-05-29
Allow instantiating vault with existing local staking contract

JakeHartnell:
Was playing around with allowing a DAO to have its own mesh vault and realized we needed to be able to instantiate the vault from the local staking contract. Thanks to @ethanfrey for feedback and figuring out how we can clean support the existing interface used in the Mesh Security UI.

### DIFF
--- a/contracts/provider/vault/src/msg.rs
+++ b/contracts/provider/vault/src/msg.rs
@@ -15,6 +15,23 @@ pub struct StakingInitInfo {
     pub label: Option<String>,
 }
 
+/// Information about the local staking contract used during vault instantiation
+/// We use untagged so older clients that just used JSON encoded StakingInitInfo
+/// will continue to work with no changes.
+#[cw_serde]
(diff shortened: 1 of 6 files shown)


contracts/provider/vault/src/msg.rs
```diff
@@ -15,6 +15,23 @@ pub struct StakingInitInfo {
...
+    Existing(ExistingContract),
+    #[serde(untagged)]
+    New(StakingInitInfo),
+}
+
+#[cw_serde]
+pub struct ExistingContract {
+    /// Address of the local staking contract
+    pub existing: String,
+}
```
JakeHartnell, line 33: It's amazing we have such super powers now. Thanks @ethanfrey for pointing this out to me.

ethanfrey (approved):
lgtm.

As minor cleanup, you could add `impl Into<LocalStakingInfo> for StakingInitInfo` but just aesthetics

PR #195, opened 2024-07-11
feat: band price feeder

### LINKED ISSUE #193: Add new price-feed data source
We utilize a single price-feed source for mesh security, which is the Osmosis TWAP price.
By extending the price-feed source, the protocol could become more decentralized, reducing dependency on osmosis. Moreover, it could enhance security by providing more data sources for price data.

The Band Protocol supports [[requesting oracle data over IBC from cw contract. So that it's possible to use band protocol to request price data for consumer contract.

I propose extending the current `remote-price-feed` contract for support query price data in band-protocol by using their cw-band : https://github.com/bandprotocol/cw-band

trinitys7:
Close #193


contracts/consumer/band-price-feed/src/bin/schema.rs
JakeHartnell, line 1: Did you mean to move this file or simply copy it?

JakeHartnell: I see other files were moved too, did you kill the old `remote-price-feed` contract?

trinitys7: I just move the schema.rs file in to correct folder. I don't remove anything.

PR #199, opened 2024-07-24
feat: transition stake

### LINKED ISSUE #89: Design clean transition from locally staked tokens to cross-staked tokens
This may well be chain-specific and definitely requires a custom module (integration) on the provider side, but we should provide a design as well as at least one sample implementation (Osmosis).

**Situation**: I have 1000 $OSMO staked normally. Now, mesh-security is deployed and I want to move them to the vault, so I can continue staking them with the same Osmosis validators, but now also leverage my collateral to cross-stake on Juno and Akash validators.

**Problem**: Since the vault only works with liquid tokens, this would require first completely unbonding the tokens (2-4 weeks), then entering the mesh-security cross-staking. If I keep the same local validators, there should be some faster "rebond-like" option.

**Idea**

One basic flow we could automate would be something like:
- Immediate unstake T tokens that user U had delegated to validator V
- Bond T tokens to the vault contract on behalf of user U
- Locally stake T tokens to validator V on behalf of user U

The end state is the same local staked validators managed by user U, just via a proxy (native-staking-proxy) contract, but same security settings (no quick exit). However, the tokens are now managed by the vault, and after this transfer, could be used for cross-staking immediately.

The above would have to be done by a custom module (as it requires more permissions than any contract can exercise), and there are probably other flows to achieve the same needs with less custom work. This issue first consists of design work, then adding a minimal implementation to either mesh-security-sdk or osmosisd

neitdung:
closes #89 
### Description
- Add `Restake` custom message to run `InstantUndelegate` function at `mesh-security-provider` module.
- Then`Deposit` coin to `vault` contract and `Delegate` through `native_staking` contract.

### Note
- Need to write new e2e test

- **Please merge #196 before this PR.**


neitdung:
> LGTM, but CI is failing. 🙃

Thank you so much, I've fixed the CI.

